PR #2, opened 2021-11-30
restructure search api

KerstenBreuer:
@deepakunni3 and me restructured the search API to:
- make the filter/facet options generic and dependent on previous search results
- ship content with the search result


KerstenBreuer:
Title for squash and merge: 
```
restructure blob fish search api
```

Description for squash and merge:
```
- make the filter/facet options generic and dependent on previous search results
- ship content with the search result
```

deepakunni3 (approved):

PR #11, opened 2021-12-10
Add count to SearchResult

deepakunni3:
- Add count to SearchResult so that it is clear what the total number of hits are, even when paginating

### DIFF
--- a/0_blob_fish/api_definitions/rest/metadata_search.yaml
+++ b/0_blob_fish/api_definitions/rest/metadata_search.yaml
@@ -90,6 +90,7 @@ components:
       properties:
         count:
           type: int
+          description: Total number of hits.
         hits:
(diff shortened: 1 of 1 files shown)


0_blob_fish/api_definitions/rest/metadata_search.yaml
```diff
@@ -88,6 +88,8 @@ components:
     SearchResult:
       type: object
       properties:
+        count:
+          type: int
```
KerstenBreuer, line 92:
```suggestion
        count:
          type: int
          description: Total number of hits.
```

KerstenBreuer (approved):
Thank you :)

deepakunni3:
Added the suggestions 👍

PR #12, opened 2022-01-24
Add API definition for Horseshoe Bat Epic

deepakunni3:
This PR adds the REST API definition(s) for Horseshoe Bat Epic


2_horseshoe_bat/technical_specification.md
```diff
@@ -0,0 +1,199 @@
...
+
+Upon successful validation, the MRS assigns a Submission ID (unique identifier) to the Submission.
+Then the Submission object, and its associated entities, are written to the metadata store (2.1).
+
+Upon a successful write operation, the Submission object (which includes a Submission ID) is
+returned to the Data Submitter for reference. If the validation fails then the data submitter
+is notified of a failure (2.2).
+
+**Note:** A Submission can have one of the following status: `in progress`, `submitted`, 
+`released`. At this stage, the Submission is neither considered 'submitted' nor
```
KerstenBreuer, line 71: Maybe we don't even need a released status because only the data set not the submission is released, right?

2_horseshoe_bat/technical_specification.md
```diff
@@ -0,0 +1,199 @@
...
+
+Upon successful completion, the MRS provides the Dataset ID to the Data
+Submitter/Data Steward (6.2).
+
+
+### 7.0 Data Steward marks a Dataset as released
+
+A Dataset is not publicly available until its status is changed to `released`.
+
+To do so, a Data Steward sends a HTTP POST request to the MRS with a Dataset ID
```
KerstenBreuer, line 162: It's a patch instead of a post isn't it?

2_horseshoe_bat/technical_specification.md
```diff
@@ -0,0 +1,199 @@
...
+## API Definitions:
+
+The definitions are hosted here:
+
+
+### RESTful/Synchronous:
+
+The RESTful service API are described using OpenAPI:
+
+- MRS:
```
KerstenBreuer, line 186: Could you insert the link to the yaml or even to the swagger UI?

2_horseshoe_bat/technical_specification.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Payload Schemas for Asynchronous Topics:
+
+The payloads for asynchronous topics are described using JSON schemas:
+
+- new_study_created: JSON Schema
```
KerstenBreuer, line 192: Same here could you insert the link?

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2404 @@
...
+        new Dataset with
+
+        the files and and link it to the given DAP ID.'
+      operationId: create_dataset_datasets_post
+      parameters:
+      - in: query
+        name: data_access_policy_id
+        required: true
+        schema:
+          title: Data Access Policy Id
```
KerstenBreuer, line 1817: Where is the data access policy created? Could we maybe embed it into the body of this request?

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2404 @@
...
+        xref:
+          description: Alternate identifiers for an entity.
+          items:
+            type: string
+          title: Xref
+          type: array
+      title: Study
+      type: object
+    Submission:
+      description: 'A grouping entity that represents information about one or more
```
KerstenBreuer, line 1383: Maybe we should create a version of this submission model that is specific for creating a new submission (posting a submission)?

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
KerstenBreuer: Same for the datasets?

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2404 @@
...
+      description: Update status of a Submission.
+      operationId: update_submission_status_submissions__submission_id__patch
+      parameters:
+      - in: path
+        name: submission_id
+        required: true
+        schema:
+          title: Submission Id
+          type: string
+      - in: query
```
KerstenBreuer, line 2290: The new status shouldn't be set via query parameter but it should be specified in the body.

KerstenBreuer (changes requested):
Awesome. Just some minor suggestions.

2_horseshoe_bat/technical_specification.md
deepakunni3: Agreed.

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2676 @@
...
+          title: Xref
+          type: array
+      title: Biospecimen
+      type: object
+    CreateDataset:
+      properties:
+        has_data_access_policy:
+          anyOf:
+          - type: string
+          - $ref: '#/components/schemas/DataAccessPolicy'
```
KerstenBreuer, line 379: Specifying a string here cannot work currently because we don't have a `POST /data_access_policy` endpoint, yet. I would suggest to remove the ambivalence here and for now and only go with the second option. If we later decide we need the additional endpoint `POST /data_access_policy`, we can still change it. (Introduce complexity when it's needed)

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2676 @@
...
+          description: The Data Access Policy that applies to this Dataset.
+          title: Has Data Access Policy
+        has_file:
+          anyOf:
+          - items:
+              type: string
+            type: array
+          - items:
+              $ref: '#/components/schemas/File'
+            type: array
```
KerstenBreuer, line 389: Same here. (Avoid ambivalence, use CreatFile schema)

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2676 @@
...
+          type: array
+      title: Dataset
+      type: object
+    DatasetStatusPatch:
+      properties:
+        id:
+          description: A unique GHGA identifier assigned to an entity for the sole
+            purpose             of referring to that entity in a global scope.
+          title: Id
+          type: string
```
KerstenBreuer, line 747: Is there a use case for updating the ID? Should we allow that at all?

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
KerstenBreuer, line 379: Moreover, when this is used to create a DataAccessPolicy, we need a special CreateDataAccessPolicy schema without the id and accession field.

KerstenBreuer (changes requested):
Great thank you. Just these few comments.

2_horseshoe_bat/technical_specification.md
```diff
@@ -0,0 +1,229 @@
...
+## API Definitions:
+
+The definitions are hosted here:
+
+
+### RESTful/Synchronous:
+
+The RESTful service API are described using OpenAPI:
+
+- MRS: [OpenAPI YAML](api_definitions/rest/metadata_repository.yaml)
```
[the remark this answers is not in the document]
deepakunni3: Yes, good idea 👍

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,2676 @@
...
+          description: Successful Response
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+          description: Validation Error
+      summary: Get a DataAccessPolicy
+  /datasets:
+    post:
```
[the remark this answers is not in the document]
deepakunni3: 👍

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
deepakunni3: Thanks for pointing that out. I think it would be much straight forward to have a `POST /data_access_policy` endpoint to support creation of a DAP. This would also over the user journey outlined in the specification. Similarly for DAC.

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
deepakunni3: 👍

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
deepakunni3: Good point. 👍

deepakunni3:
@KerstenBreuer Thanks for the review. I have incorporated your suggestions.

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,3886 @@
...
+        description:
+          description: Description of an entity.
+          title: Description
+          type: string
+        ega_accession:
+          description: A unique European Genome-Phenome Archive (EGA) identifier             assigned
+            to an entity for the sole purpose of referring to that entity             within
+            the EGA federated network.
+          title: Ega Accession
+          type: string
```
KerstenBreuer: The EGA accession is not assigned by the user but by our system.

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,3886 @@
...
+        description:
+          description: A short description for the Data Access Policy.
+          title: Description
+          type: string
+        ega_accession:
+          description: A unique European Genome-Phenome Archive (EGA) identifier             assigned
+            to an entity for the sole purpose of referring to that entity             within
+            the EGA federated network.
+          title: Ega Accession
+          type: string
```
KerstenBreuer: The EGA accession is not assigned by the user but by our system.

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
```diff
@@ -0,0 +1,3886 @@
...
+          description: The Data Access Policy that applies to this Dataset.
+          title: Has Data Access Policy
+        has_file:
+          anyOf:
+          - items:
+              type: string
+            type: array
+          - items:
+              $ref: '#/components/schemas/CreateFile'
+            type: array
```
KerstenBreuer: Files should only be created using submission objects, right?
Should we really allow to post new Files here?
Probably it's good to just allow an ID to an already existing file here. Don't you agree?

KerstenBreuer (changes requested):
Cool, I think we are almost there. Just these few things.

2_horseshoe_bat/api_definitions/rest/metadata_repository.yaml
deepakunni3: Good catch! Totally missed that 👍

deepakunni3:
Thanks for the review @KerstenBreuer ! I think its ready 👍

KerstenBreuer (approved):

PR #19, opened 2022-05-03
Added Multipart Upload Specifications

MoritzHahn1337:
Added additional multipart upload specifications to epic 4: star-nosed mole.


MoritzHahn1337:
@KerstenBreuer 3 questions:

1. Should I leave it in an extra file to preserve history or merge with the original?
2. Is this according to what we discussed earlier?
3. Is this too detailed and should I move the inner workings of the API endpoints to the respective task?

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Scope:
+A scope definition can be found here: https://wiki.verbis.dkfz.de/x/fAFzCQ
+
+## User Journeys
+
+The addition to this epic covers the following user journeys:
+
+### Multipart Data Upload:
+![Data Upload](./images/multipart_data_upload.png)
```
KerstenBreuer, line 13:
```suggestion
![Data Upload](./images/multipart_data_upload.jpg)
```

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+
+A Data Submitter specifies the file ID, the URL to the Upload Contoller API, and the file path on the local file system using the CLI interface of the client (1.0). Internally, the CLI client translates the user-defined data into a request to the Upload Contoller API to obtain an S3 upload id (1.1). The client sends a request to the Upload Controller for each individual part of the uploaded file and recieves a presigned post for each individual part (1.2). The client reads multiple parts from the source file and uploads them as a stream using the pre-signed post (1.3). Once the upload has been completed the client sends a confirmation to the Upload Controller API (1.4).
```
KerstenBreuer, line 15:
```suggestion
A Data Submitter specifies the file ID, the URL to the Upload Contoller API, and the file path on the local file system using the CLI interface of the client (1.0). Internally, the CLI client translates the user-defined data into a request to the Upload Contoller API to obtain an upload id (1.1). The client sends a request to the Upload Controller for each individual part of the uploaded file and recieves a presigned post for each individual part (1.2). The client reads multiple parts from the source file and uploads them as a stream using the pre-signed post (1.3). Once the upload has been completed the client sends a confirmation to the Upload Controller API thereby specifying the eTag for each file part (1.4).
```

We can just stay generic and say there is an upload ID, we might want to only keep the actual ID we get from S3 private in our database and expose another ID to the public.

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+
+## UCS:
+
+The following Updates are performed in the Upload Controller Service:
+
+### Database:
+
+A new table "multipart_uploads" is added, containing the following columns:
+```
+- s3_upload_id (Primary key)
```
KerstenBreuer, line 29: Maybe rather:
```suggestion
- upload_id (Primary Key, UUID)
- s3_upload_id (unique)
```

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+- file_id (not unique)
+- upload_status
+```
+
+file_state is an Enum and can have the following values:
```
KerstenBreuer, line 34:
```suggestion
Thereby, the upload_status is an Enum and can have the following values:
```

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+```
+- pending
+- canceled
+- failed
+- uploaded
+- accepted
+- rejected
+```
+
+The column "upload_status" from then table "files" is removed.
```
KerstenBreuer, line 45:
```suggestion
Please note, not more that one upload per file may have a state that is set to `pending` , `uploaded`, or `accepted`. Morover, within the list of states from uploads corresponding to one file, these `pending` , `uploaded`, and `accepted` are mutually exclusive.

This also means that once an upload attempt corresponding to a file is set to `accepted`, no new uploads can be created for that file. In a future epic, we will implement a mechanism that allows controlled re-upload of a file by explicitly requesting to depreciate an old upload. We might also consider an `is_open` flag that is specified per file to control whether new upload attempts are currently allowed for that file.

The column "upload_status" from the table "files" is removed.
```

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+A column "s3_upload_id" (unique) is added to the table "files".
```
KerstenBreuer, line 46: I think this is not needed, because the "current" upload attempt can be reconstructed from the uploads table.

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+with:
+```
+POST /files/{file_id}/multipart_uploads
+    if there is already a multipart_upload with file_id and upload_status == pending, updated or accepted
+        returns error
+    else:
+        gets s3_upload_id from S3
+        creates new row in multipart_uploads with file_id, s3_upload_id and upload_status=pending
+        put s3_upload_id into database
+returns s3_upload_id and the part size in bytes for this download (default: 16 MiB)
```
KerstenBreuer, line 69:
```suggestion
returns upload_id and the part size in bytes for this download (default: 16 MiB)
```

KerstenBreuer: We don't know the exact size of the file before we upload right? Could we thus give instructions on the part size that shall be used?

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+Add:
+```
+GET /files/{file_id}/multipart_uploads?status=pending
+returns the s3_upload_id in the table multipart_uploads where file_id=file_id and upload_status=pending (There should never be more than one.) or None
```
KerstenBreuer, line 75:
```suggestion
returns the upload_id in the table multipart_uploads where file_id=file_id and upload_status=pending (There should never be more than one.) or None
```

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+Add:
+```
+POST /multipart_uploads/{upload_id}/part/{part_no}/presigned_post
```
KerstenBreuer, line 80: Should also be plural for REST-compliance.
We might also consider to rename it to simple `.../presigned_urls` or `.../urls`.

```suggestion
POST /multipart_uploads/{upload_id}/part/{part_no}/presigned_posts
```

4-star-nosed_mole/multipart_upload_specifications.md
```diff
@@ -0,0 +1,108 @@
...
+    body: state: UploadState (Uploaded)
+    sets upload_status to uploaded
+returns: 204
+```
+
+with:
+```
+PATCH /multipart_uploads/{upload_id}/
+    body:       status==uploaded/cancelled
+                part_no to eTag mapping (only if status==uploaded)
```
KerstenBreuer, line 97: Maybe we can also store that mapping in the database?

KerstenBreuer (changes requested):
Nice!

Regarding your questions:
1. Maybe it would be good to just integrate it (or append) to the existing technical_specification.md
2. Yes that's pretty much it 👍 
3. I don't see a problem with these details.

Generally, I think it would be good to translate these endpoint descriptions into an OpenAPI doc. That would make it even more clear how request and response bodies would look. But maybe we can do that in a separate PR. It might also make sense that Thomas is doing that since he will implement that API in the UCS. What do you think?

4-star-nosed_mole/multipart_upload_specifications.md
MoritzHahn1337: The CLI needs the s3_upload_id to upload to s3.

4-star-nosed_mole/multipart_upload_specifications.md
MoritzHahn1337: I assumed, we get those information from the metadata service

4-star-nosed_mole/multipart_upload_specifications.md
MoritzHahn1337: But what we recieve are in fact not urls, but an url with additional information we have to send in the post body, so url doesn't fit here.

4-star-nosed_mole/multipart_upload_specifications.md
MoritzHahn1337: Why would we need to?

MoritzHahn1337:
> Generally, I think it would be good to translate these endpoint descriptions into an OpenAPI doc. That would make it even more clear how request and response bodies would look.

Using a pre-created OpenAPI doc would kind of defeat the purpose of using FastAPI for the Endpoints in the first place.

KerstenBreuer (approved):

PR #26, opened 2022-05-30
Change status to upload_status, add a 400 return for the patch endpoint

MoritzHahn1337:
Using status in a HTML GET url produces errors.

Status has to be changed to upload_status everywhere
Added the possibility for a 400 error

### DIFF
--- a/4-star-nosed_mole/api_definitions/rest/ucs.yaml
+++ b/4-star-nosed_mole/api_definitions/rest/ucs.yaml
@@ -101,8 +101,8 @@ paths:
           description: Multi-part upload successfully updated.
         "400":
           description: |
-            The multi-part upload with the given ID can't be
-            set to this status.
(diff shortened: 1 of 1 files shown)


4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -99,6 +99,10 @@ paths:
       responses:
         "204":
           description: Multi-part upload successfully updated.
+        "400":
+          description: |
+            The multi-part upload with the given ID can't be
+            set to this status.
```
mephenor, line 105:
```suggestion
            Cannot set invalid status for multi-part upload
            with the given ID.
```
Should make it a bit more clear that the status provided is wrong

mephenor (approved):
Small change to wording should make the intention behind the error more clear without having to see the error code

PR #27, opened 2022-06-07
[star-nosed mole] general improvements and bugfixes to the UCS spec

KerstenBreuer:


4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -101,20 +101,19 @@ paths:
           description: Multi-part upload successfully updated.
         "400":
           description: |
-            Cannot set invalid status for multi-part upload
-            with the given ID.
+            The provided status value was invalid.
```
MoritzHahn1337, line 104: We probably should further distinguish here.

This description just says, that we cannot change to the specific status, that was specified.
However, it could also be that the status currently or permanently can't be changed.

```suggestion
            The provided status value was invalid or the status of this multi-part upload currently or permanently can't be changed.
```

MoritzHahn1337 (approved):

PR #28, opened 2022-06-07
add pre-sigend url response schema

MoritzHahn1337:


4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -131,12 +131,12 @@ paths:
       responses:
         "200":
           description: |
-            The newly create pre-signed POST.
+            The newly create pre-signed url.
```
mephenor, line 134:
```suggestion
            The newly created pre-signed url.
```

mephenor (approved):

KerstenBreuer (approved):

4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -163,6 +162,15 @@ components:
...
 
+    PreSignedURL:
+      type: object
+      properties:
+        presigned_url:
+          type: string
+          description: The pre-signed URL for uploading the file part
+      required:
+        - presigned_url
+
```
KerstenBreuer, line 173: It would be nice if we could use the same structure as the DRS spec:
```
    AccessURL:
      type: object
      required:
        - url
      properties:
        url:
          type: string
          description: >-
            A fully resolvable URL that can be used to fetch the actual object
            bytes.
        headers:
          type: array
          items:
            type: string
          description: >-
            An optional list of headers to include in the HTTP request to `url`.
            These headers can be used to provide auth tokens required to fetch
            the object bytes.
          example: 'Authorization: Basic Z2E0Z2g6ZHJz'
```
https://editor.swagger.io/?url=https://ga4gh.github.io/data-repository-service-schemas/preview/develop/openapi.yaml

Since the headers are not used in our case, we can simplify it to:
```suggestion
    AccessURL:
      type: object
      required:
        - url
      properties:
        url:
          type: string
          description: >-
            A fully resolvable URL that can be used to fetch the actual object
            bytes.
```

KerstenBreuer (changes requested):
Maybe we can

KerstenBreuer (approved):

PR #29, opened 2022-06-10
Feature/exception ids gdev 954

mephenor:
Summary for squash & merge:
```
Added a mermaid diagram to visualize possible state changes
Added exception IDs to error code response descriptions 
```

Please check if naming of the exception IDs is okay.
For most I tried to follow a pattern where I appendend a Suffix to describe the path/action affected.

For the easy ones, i.e. one exception ID only, I left the existing description.
For status changes I added exception IDs for all meaningful ones, though I am not sure if I missed some and if we need to actually differentiate between all of the ones I wrote

### DIFF
--- a/4-star-nosed_mole/api_definitions/rest/ucs.yaml
+++ b/4-star-nosed_mole/api_definitions/rest/ucs.yaml
@@ -173,7 +173,7 @@ paths:
         "404":
           description: |
             Exceptions by ID:
-              - uploadNotExistingPresignPart: The multi-part upload with the given ID
+              - noSuchUpload: The multi-part upload with the given ID
                 does not exist.
 
 components:
--- a/4-star-nosed_mole/api_definitions/rest/ucs.yaml
+++ b/4-star-nosed_mole/api_definitions/rest/ucs.yaml
@@ -24,10 +24,14 @@ paths:
                 $ref: "#/components/schemas/FileMetadata"
         "403":
           description: |
-            The user is not registered as a Data Submitter for the
-            corresponding file.
+            Exceptions by ID:
+              - userNotRegisteredFileAccess: The user is not registered
+                as a Data Submitter for the corresponding file.
         "404":
-          description: The file with the given ID has not (yet) been registered for upload.
+          description: |
+            Exceptions by ID:
+             - fileNotRegisteredAccess: The file with the given ID has
+               not (yet) been registered for upload.
 
   /uploads:
     post:
@@ -47,14 +51,18 @@ paths:
                 $ref: "#/components/schemas/UploadDetails"
         "400":
           description: |
+            Exceptions by ID:
+              - TODO:
+              -
             It is currently not possible to create a new upload for
             the file with the specified ID because another upload for that
             file is already active or has been accepted
             (its status is not "failed", "cancelled", or "rejected").
         "403":
           description: |
-            The user is not registered as a Data Submitter for the
-            corresponding file.
+            Exceptions by ID:
+              - userNotRegisteredUploadCreation: The user is not registered
+                as a Data Submitter for the corresponding file.
 
   /uploads/{upload_id}:
     get:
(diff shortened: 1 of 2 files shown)


4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -24,10 +24,14 @@ paths:
...
-            The user is not registered as a Data Submitter for the
-            corresponding file.
+            Exceptions by ID:
+              - userNotRegisteredFileAccess: The user is not registered
+                as a Data Submitter for the corresponding file.
         "404":
-          description: The file with the given ID has not (yet) been registered for upload.
+          description: |
+            Exceptions by ID:
+             - fileNotRegisteredAccess: The file with the given ID has
```
KerstenBreuer, line 33:
```suggestion
             - fileNotRegisteredYet: The file with the given ID has
```

4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -101,13 +116,27 @@ paths:
...
+              - invalidChangeFromAccepted: Upload has already been accepted, a status change or
+                new upload attempt is no longer possible.
+              - invalidChangeFromRejected: Cannot change status from rejected for this upload attempt.
+                If you want to restart, initiate a new multi-part upload instead.
         "403":
           description: |
-            The user is not registered as a Data Submitter for the
-            file corresponding to this multi-part upload.
+            Exceptions by ID:
+              - userNotRegisteredUploadPatch: The user is not registered as a
```
KerstenBreuer, line 133:
```suggestion
              - noFileAccess: The user is not registered
```

4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -101,13 +116,27 @@ paths:
...
+                Data Submitter for the file corresponding to this multi-part upload.
         "404":
-          description: The multi-part upload with the given ID does not exist.
+          description: |
+            Exceptions by ID:
+              - uploadNotExistingPatch: The multi-part upload with the given ID
```
KerstenBreuer, line 138: Can be the same as above:

```suggestion
              - noSuchUpload: The multi-part upload with the given ID
```

KerstenBreuer, line 138: Can be the same as above:

```suggestion
              - noSuchUpload: The multi-part upload with the given ID
```

4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -138,10 +167,14 @@ paths:
...
-            The user is not registered as a Data Submitter for the
-            file corresponding to this multi-part upload.
+            Exceptions by ID:
+              - userNotRegisteredPresignPart: The user is not registered as a
+                Data Submitter for the file corresponding to this multi-part upload.
         "404":
-          description: The multi-part upload with the given ID does not exist.
+          description: |
+            Exceptions by ID:
+              - uploadNotExistingPresignPart: The multi-part upload with the given ID
```
KerstenBreuer, line 176: Can be the same as above:

```suggestion
              - noSuchUpload: The multi-part upload with the given ID
```

KerstenBreuer, line 176: Can be the same as above:

```suggestion
              - noSuchUpload: The multi-part upload with the given ID
```

KerstenBreuer (changes requested):
Very nice, just a couple of suggestions for renaming.
Please let me know what you think?

4-star-nosed_mole/api_definitions/rest/ucs.yaml
mephenor: Not sure about this one. There is no way to have knowledge if the registration is coming/ongoing at that point in the code, right? Then it might be better to just have that as `fileNoteRegisterd`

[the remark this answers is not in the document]
mephenor: Github did a weird thing here... Seems like this should have been part of the previous one.
Does this mean two different Exception IDs?

KerstenBreuer: Yes that makes sense

KerstenBreuer (approved):

PR #34, opened 2022-06-29
Add fileNotRegistered exceptionId for /uploads endpoint

MoritzHahn1337:
Description:
```
Added fileNotRegistered exceptionID added as a 404 error for POST: /uploads endpoint
```

### DIFF
--- a/4-star-nosed_mole/api_definitions/rest/ucs.yaml
+++ b/4-star-nosed_mole/api_definitions/rest/ucs.yaml
@@ -61,16 +61,13 @@ paths:
               - uploadAttemptPresentAccepted: Imposible to create a new upload for
                 the file with the specific ID. Another upload in accepted status
                 already exists for this file.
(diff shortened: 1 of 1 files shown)


4-star-nosed_mole/api_definitions/rest/ucs.yaml
```diff
@@ -66,6 +66,11 @@ paths:
             Exceptions by ID:
               - noFileAccess: The user is not registered as a Data Submitter
                 for the corresponding file.
+        "404":
```
KerstenBreuer, line 69: Good idea, I am not sure about the status code, though. I think typically the 404 is used when the requested URL does not exist. Maybe it is better to also move this to the 400-list?

KerstenBreuer (changes requested):

KerstenBreuer (approved):

PR #40, opened 2022-07-06
eliminate abiguity of the DRS spec (GDEV-1030)

KerstenBreuer:
by removing unused options and making all used fields required.

### DIFF
--- a/4-star-nosed_mole/api_definitions/rest/drs3.yaml
+++ b/4-star-nosed_mole/api_definitions/rest/drs3.yaml
@@ -1,0 +1,225 @@
+openapi: 3.0.1
+info:
+  title: Data Repository Service on S3 (DRS3)
+  description: |
+    A service managing access to file objects stored on an S3-compatible Object Storage.
+
+    This is an implementation of the DRS standard from the Global Alliance for Genomics
+    and Health, please find more information at:
+    [https://github.com/ga4gh/data-repository-service-schemas](https://github.com/ga4gh/data-repository-service-schemas)
+  termsOfService: https://www.ga4gh.org/terms-and-conditions/
+  license:
+    name: Apache 2.0
+    url: https://raw.githubusercontent.com/ga4gh/data-repository-service-schemas/master/LICENSE
+  version: 1.1.0
+servers:
+  - url: /ga4gh/drs/v1
+security:
+  - {}
+  - authToken: []
+tags:
+  - name: DataRepositoryService
+paths:
+  /objects/{object_id}:
+    get:
+      tags:
+        - DataRepositoryService
+      summary: Get info about a `DrsObject`.
+      description:
+        Returns object metadata, and a list of access methods that can
+        be used to fetch object bytes.
+      operationId: GetObject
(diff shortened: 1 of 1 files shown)


KerstenBreuer:
Hi @mephenor, since Moritz is not available, maybe you can have a brief look?

KerstenBreuer:
This has been adapted from the DRS spec at https://editor.swagger.io/?url=https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.1.0/swagger.yaml

4-star-nosed_mole/api_definitions/rest/drs3.yaml
```diff
@@ -0,0 +1,225 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        500:
+          description: An unexpected error occurred.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+      x-swagger-router-controller: ga4gh.drs.server
```
mephenor, line 86: What's the role of `x-swagger-router-controller`?

4-star-nosed_mole/api_definitions/rest/drs3.yaml
```diff
@@ -0,0 +1,225 @@
...
+      required:
+        - checksums
+        - created_time
+        - id
+        - self_uri
+        - size
+        - access_methods
+        - name
+      type: object
+      properties:
```
mephenor, line 121: There are some properties missing defined in the drs spec:
updated_time and version make no sense for our use case, I guess.
What about aliases, description and mime_type?

4-star-nosed_mole/api_definitions/rest/drs3.yaml
```diff
@@ -0,0 +1,225 @@
...
+        url:
+          type: string
+          description:
+            A fully resolvable URL that can be used to fetch the actual
+            object bytes.
+        headers:
+          type: array
+          description:
+            An optional list of headers to include in the HTTP request
```
mephenor, line 187:
```suggestion
            A list of headers to include in the HTTP request
```
As it is required in our case

mephenor (approved):
As far as I can grasp this, it should be good.
I just have two questions a minor correction

4-star-nosed_mole/api_definitions/rest/drs3.yaml
KerstenBreuer: That was also inherited from the original DRS spec.

But I don't think we will have immediate use for it:
https://github.com/swagger-api/swagger-node/blob/master/docs/controllers.md

Will remove it.

4-star-nosed_mole/api_definitions/rest/drs3.yaml
KerstenBreuer: We currently don't use them either. All of these information would be queriable via our metadata services. It might be a bit redundant and create extra work to repeat it here. But at a later stage we could re-evaluate.

PR #41, opened 2022-07-06
use httpyexpect-style exceptions in DRS3 (GDEV-1012)

KerstenBreuer:
thereby also removed unused response codes from the spec.


KerstenBreuer:
Builds up on top of #40.

mephenor (approved):
Should be in line with ucs.yaml custom exception.

There's also https://github.com/ghga-de/drs3/blob/main/drs3/api/openapi.yaml.
That's superseded by this document, right?

PR #56, opened 2022-11-04
Add tech spec for Red Knob Sea Star

Cito:


20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
```diff
@@ -0,0 +1,367 @@
...
+          schema:
+            title: Internal user ID
+            type: string
+      responses:
+        '204':
+          description: User claims were successfully deleted.
+        '404':
+          description: The user was not found.
+        '422':
+          description: Validation error in submitted user ID.
```
KerstenBreuer, line 106: Should claims ever be deleted? Or just let them expire or be revoked?

KerstenBreuer: If we need to delete them, do we have a use case for this "delete all" endpoint or would the single delete endpoint be enough?

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
```diff
@@ -0,0 +1,367 @@
...
+          required: true
+          schema:
+            title: Internal claim ID
+            type: string
+        - name: revocation_date
+          in: query
+          required: true
+          schema:
+            title: Revocation Date
+            type: string
```
KerstenBreuer, line 161: This should go into the body as it is not used for resource identification but is the data to patch.

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
```diff
@@ -0,0 +1,367 @@
...
+          minLength: 1
+          type: string
+          format: uri
+        sub_source:
+          title: Asserting sub-organization
+          maxLength: 2083
+          minLength: 1
+          type: string
+          format: uri
+          nullable: true
```
KerstenBreuer, line 238: Would this be the identifier of the DAC making the assertion?

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
```diff
@@ -0,0 +1,367 @@
...
+            items:
+              $ref: '#/components/schemas/Condition'
+          nullable: true
+        id:
+          title: ID
+          type: string
+          description: Internally used ID of the claim
+      description: A claim made about a user with a claim ID
+
+    ClaimData:
```
KerstenBreuer, line 258: Maybe:
```suggestion
    ClaimCreation:
```

KerstenBreuer (dismissed):
Maybe some minor points to discuss.

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
Cito: May be useful if a user is deleted from the database. But of course we can also get all claims for the user and then remove them individually.

Cito: Claims can conain personal data about the user, so we may not have a legal basis to store them forever.

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
Cito: Yes, that was meant to go into the body. Changed in d0197ce5d142b3241abf4af5aa7e9d94d7946d66.

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
Cito: Yes, or maybe the hub if it is an assertion made by us. The visa only have a single source attribute, which then would contain both the organization and the sub-organization (e.g. by using a subdomain), or contain only one of them. But in the repository it may be good to have them separately.

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
Cito: Ok, changed that to better match the terminology of hexkit.

20-red-knob-sea-star/api_definitions/rest/claims_repository.yaml
Cito: Actually the user claims could also be deleted automatically if the user is deleted (via events or database cleanup jobs), so we don't really need that.

The deletion of individual claims may be needed to correct mistakes. I think the instance that is able to create a claim should also be able to delete it. Proper auditing of such operations is a separate concern that can be handled elsewhere.

Cito: So I have removed the endpoint to delete all claims for the user now, but I would like to keep the other one.

KerstenBreuer (approved):

PR #69, opened 2023-04-05
Add technical specification for Notification Service (Humphead Wrasse)

TheByronHimes:


KerstenBreuer (changes requested):
Very nice. Maybe mention that emails are send by interacting with an SMTP server. Moreover, add config parameters of this microservice such as smpt config and the jinja2-based HTML template.

KerstenBreuer (approved):

PR #70, opened 2023-04-13
Added Purple Frog specs

mephenor:


30-purple_frog/technical_specification.md
```diff
@@ -0,0 +1,32 @@
...
+The first call replaces an existing function that currently gets its information from the config/environment variables.
+API specs for the work package service endpoints can be found here: https://github.com/ghga-de/work-package-service/blob/main/openapi.yaml
+
+
+### Work Package Access Token processing
+
+The work package access token is an alphanumeric string that needs to be processed by the connector.
+Due to security considerations the token should not be provided as command line option, as this would be included in the command line history.
+Instead, the user should be asked for the token when needed.
+This token is encrypted and needs to be decrypted before use and needs to be provided when calling the WPS to access package information and retrieve work order tokens.
```
KerstenBreuer, line 22: Please consult on this with Christoph if not already done. He used a non-crypt4GH-based encryption method.

KerstenBreuer (approved):

30-purple_frog/technical_specification.md
Cito: Actually it is the same method/library that is also used by the crypt4GH library under the hood. I use it directly because we only want to encrypt with the user's key, not sign with a key of ours at the same time as the crypt4GH library does.

But yes, @mephenor let's discuss this point tomorrow.

PR #74, opened 2023-04-21
Update spec for work package service (GSI-25)

Cito:
Made some changes to the specs of the work package service (Dracula Ant epic).

Note that some things that still need discussion and clarification are marked as *TODO*.


26-dracula-ant/technical_specification.md
```diff
@@ -22,18 +22,20 @@ Used by the web frontend to create work packages:
...
-    - `public_key`: string (the user's public Crypt4GH key)
+    - `file_ids`: array of strings  (null = all files of the dataset)
+    - `user_public_crypt4gh_key`: string (the user's public Crypt4GH key)
   - response body:
-    - `id`: str (the work package ID)
-    - `token`: str (encrypted work package access token)
+    - `id`: str (the ID of the created work package)
+    - `token`: str (encrypted work and base64 encoded package access token)
+
+*TODO:* The response body could also contain info about the expiration of the token.
```
KerstenBreuer, line 31: That's a good idea, for now we can just use the lifetime of the work package. We don't need to consider the expiration of the controlled access grant. However, that might come in the future.

26-dracula-ant/technical_specification.md
```diff
@@ -55,38 +57,59 @@ Deletion of work packages should be logged or implemented as deactivation (immed
...
-The frontend should display the work package ID together with the encrypted work package access token separated by a colon as a single string that can be copied and pasted to the CLI client, so that the CLI client gets both pieces of information.
+The frontend should display the work package ID together with the encrypted work package access token separated by a colon as a single string that can be copied and pasted to the CLI client (GHGA connector), so that the CLI client gets both pieces of information.
 
 ### Work order token
 
 This is a JWT with a very short time to live (maximum 30 seconds). It is signed by the Work Package Service and then encrypted with the `public_key` of the user, similar to the work package access token. It contains the following claims:
 
 - `type`: `download` or `upload`
 - `file_id`: the ID of the file that shall be downloaded or uploaded
+- `file_ext`: the file extension of the requested filed including the dot
```
KerstenBreuer, line 68: I think this should rather be a part of the work package. The information is available from the beginning on. It does not change between work orders.

26-dracula-ant/technical_specification.md
```diff
@@ -55,38 +57,59 @@ Deletion of work packages should be logged or implemented as deactivation (immed
...
+- `GET /download-access/users/{user_id}/datasets`
+  - authorization: only internal from download controller (via Istio)
+  - returns a list of all dataset IDs that can be downloaded by the user
+
+- `GET /download-access/users/{user_id}/datasets/{dataset_id}`
   - authorization: only internal from download controller (via Istio)
   - returns `true` or `false` as a scalar resource
 
-In order to facilitate authorization, the path of this endpoint starts with `datasets` and not with `users` which is already used by other endpoints fo the claims repository.
+*TODO:* Instead of true/false, the latter endpoint could also return null or the date when the access expires (null could be confused with unlimited access though). This could be used by the Work Package Service to limit the expiration date of the tokens, even though the Work Package Service re-checks access on every operation anyway.
```
KerstenBreuer, line 92: I think this is a good idea but maybe not that essential now. Should we do this as part of another work package?

26-dracula-ant/technical_specification.md
```diff
@@ -55,38 +57,59 @@ Deletion of work packages should be logged or implemented as deactivation (immed
...
-In order to check whether a given user has *upload* access to a given dataset, we may introduce a similar endpoint that will be based on checking custom visa types for dataset submission in the claims repository. Alternatively, we could store a list of internal user IDs of dataset submitters also in the `DatasetFiles` collection described below and update them similarly to the file IDs belonging to the dataset. (To be clarified, will be implemented later.)
+#### Access checks by the Frontend
+
+In order to provide the user with a list of downloadable datasets, the Work Package Service combines the information from the access checks explained above with information about the dataset as explained below and makes it avaliable to the frontend via the following endpoint:
+
+- `GET /datasets`
+  - auth header: internal access token with the user context
+  - returns a list of all dataset IDs that can be downloaded by the user
+
+*TODO*: Maybe this should be `/users/{user_id}/datasets` to make it more RESTful? However, the user registry has the same prefix, and the API gateway can only use the prefix to map requests to services. Any better idea?
```
KerstenBreuer, line 110: Yes that sounds good.

KerstenBreuer: Each service should have a unique prefix either way

26-dracula-ant/technical_specification.md
```diff
@@ -55,38 +57,59 @@ Deletion of work packages should be logged or implemented as deactivation (immed
...
+
+*TODO*: Maybe the dataset objects returned here could also be supplemented with information on when user access expires, obtained via the acess checks explained above (if these are extended to provide this information). This information could then be also shown to the user on the profile page or the work package creation page.
```
KerstenBreuer, line 112: A good extension for later Epics.

26-dracula-ant/technical_specification.md
KerstenBreuer: Just saw that it already is part of it. Can be deleted from the work order token.

26-dracula-ant/technical_specification.md
```diff
@@ -99,31 +122,38 @@ The database of the Work Package Service should store the following WorkPackage
...
+The `files` mapping must not be empty. It contains either the `file_ids` specified by the user when creating the work package (in this case the Work Package Service checks whether they belong to the given `dataset_id`), or the full list of all `file_id`s that belong to the given `dataset_id` if the user didn't specify a list of file IDs. The values of this mapping are the file extensions including the dot.
 
 An entry in the `WorkPackage` collection is only created by the Work Package Service after verification that the user with the given `user_id` is allowed to access the dataset with the given `dataset_id` (see access checks above).
 
-In the database, the Work Package Service also keeps track of which file IDs belong to which dataset and which users are allowed to submit data for the individual datasets. The following association collection is used for this purpose:
+In the database, the Work Package Service also keeps track of datasets, particularly which file IDs belong to which dataset, how the datasets and files are named, (and maybe also which users are allowed to submit data for the individual datasets). The following association collection is used for this purpose:
```
KerstenBreuer, line 141: Filenames are not shared with the requester. I would not duplicate permission information. The claims repo and tge visa issuer should be responsible for that.

26-dracula-ant/technical_specification.md
```diff
@@ -99,31 +122,38 @@ The database of the Work Package Service should store the following WorkPackage
...
-  file_ids: list[str]  # all file IDs that are part of the dataset
+Dataset:
+  id: str  # id of the dataset
+  title: str  # short title of the dataset
+  description: str  # long description of the dataset
+  files: list[DatasetFile]  # all files contained in the dataset
+
+DatasetFile:
+  id: str  # id of the file
+  description: str  # a description of the file
```
KerstenBreuer, line 152: Why is this needed?

KerstenBreuer (changes requested):
Really nice. Almost a documentation for the work package service.

26-dracula-ant/technical_specification.md
Cito: I thought the DCS might need it to deliver the proper content type by that should be handled by S3 already. So yes, I removed from the work order token.

26-dracula-ant/technical_specification.md
```diff
@@ -55,38 +57,59 @@ Deletion of work packages should be logged or implemented as deactivation (immed
...
 ### Work order token
 
 This is a JWT with a very short time to live (maximum 30 seconds). It is signed by the Work Package Service and then encrypted with the `public_key` of the user, similar to the work package access token. It contains the following claims:
 
 - `type`: `download` or `upload`
 - `file_id`: the ID of the file that shall be downloaded or uploaded
+- `file_ext`: the file extension of the requested filed including the dot
 - `user_id`: the internal ID of the user
-- `public_key`: the public key stored in the work package
+- `public_key`: the public key of the user stored in the work package
```
MoritzHahn1337, line 70:
```suggestion
- user_public_crypt4gh_key`: the public key of the user stored in the work package
```

maybe rename this for consistenxy with the work package

26-dracula-ant/technical_specification.md
```diff
@@ -22,18 +22,20 @@ Used by the web frontend to create work packages:
   - request body:
     - `dataset_id`: str (the ID of a dataset)
     - `type`: enum (download/upload)
-    - `file_ids`: array of strings  (empty = all files of the dataset)
-    - `public_key`: string (the user's public Crypt4GH key)
+    - `file_ids`: array of strings  (null = all files of the dataset)
+    - `user_public_crypt4gh_key`: string (the user's public Crypt4GH key)
```
MoritzHahn1337, line 26:
```suggestion
    - `user_public_crypt4gh_key`: str (the user's public Crypt4GH key)
```

consistency

MoritzHahn1337 (approved):

26-dracula-ant/technical_specification.md
Cito: > Yes that sounds good.

If our public API has the prefix `api` and the work package service has its own additional prefix `wps`, then the URL for getting work-packages would be `/api/wps/work-packages`. I don't really like this since it contains duplication and makes the implementation details (services) publicly visible instead of providing an implementation-independent API - `/api/work-packages` would be nicer.

26-dracula-ant/technical_specification.md
```diff
@@ -99,31 +122,38 @@ The database of the Work Package Service should store the following WorkPackage
...
-  email: str  # he email address of the user
+  email: str  # the email address of the user
+  user_public_crypt4gh_key: str  # the user's public Crypt4GH key
   token_hash: str  # hash of the work package access token
   created: datetime  # creation date of this work package
   expires: datetime  # expiry date of this work package
 ```
 
-The `file_ids` list must not be empty. It contains either the `file_ids` specified by the user when creating the work package (in this case the Work Package Service checks whether they belong to the given `dataset_id`), or the full list of all `file_id`s that belong to the given `dataset_id`.
+The `files` mapping must not be empty. It contains either the `file_ids` specified by the user when creating the work package (in this case the Work Package Service checks whether they belong to the given `dataset_id`), or the full list of all `file_id`s that belong to the given `dataset_id` if the user didn't specify a list of file IDs. The values of this mapping are the file extensions including the dot.
```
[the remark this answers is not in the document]
Cito: Exactly.

26-dracula-ant/technical_specification.md
Cito: Ok, formulated that it a bit more clearly.

I'm not sure if access permission to *submit* data should also be part of the claims repo. Do you think so? (Then at least I would use a custom visa type for that.)

26-dracula-ant/technical_specification.md
Cito: It's needed if we (ever) add an alternative UI for selecting files manually instead of providing a list of file IDs. But yes, maybe a case of YAGNI.

26-dracula-ant/technical_specification.md
Cito: The idea was to keep the work order token small (it's passed in the header). Usually JWT claim names are very short.

But yes, consistency is probably more important. I renamed it now.

26-dracula-ant/technical_specification.md
Cito: Removed it now (also in the WPS).

26-dracula-ant/technical_specification.md
mephenor: What would the advantage of using a date be compared to just checking, i.e. what's the context here?

mephenor (dismissed):

26-dracula-ant/technical_specification.md
Cito: @mephenor The advantage would be that the WPS could use this as upper limit for the expiration date of the work package, and that the expiration date could also be communicated to the users in the frontend.

26-dracula-ant/technical_specification.md
KerstenBreuer: We don't have one single integrated API landscape at the moment and we didn't plan to do so. The prefixes can be seen as namespaces or partitions of the API. This allows us to think of the same resources in different contexts in a different way. This is not an uncommon practice, the GA4GH is promoting a similar pattern. E.g. for getting a file via DRS you need to contact https://drs.example.org/ga4gh/drs/v1/objects/{object_id}. There even the version of the service API is visible. If you like, we can put onto the refactoring list to discuss the pros and cons of that approach. However, for now, we can use the flexibility gained by the namespacing. Thus multiple services can use the same resources in their endpoints.

26-dracula-ant/technical_specification.md
KerstenBreuer: Should we remove this and put it onto the refactoring list?

26-dracula-ant/technical_specification.md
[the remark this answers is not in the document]
KerstenBreuer: OK, maybe, we can rephrase it a little to make this clear.

26-dracula-ant/technical_specification.md
KerstenBreuer: That might be a good implementation. However, we discuss that as part of a different epic.

KerstenBreuer (dismissed):

Cito:
Hopefully addressed all the comments now. Mentioned the extensions that can be addressed in a future epic and put that in the refactoring list.

KerstenBreuer (approved):

PR #80, opened 2023-05-30
add tech spec for beluga whale epic

dontseyit:


35-beluga-whale/technical_specification.md
```diff
@@ -0,0 +1,56 @@
...
+- Setting the initial state for the file to be downloaded:
+    - Informing the Internal File Registry Service about the file stored permanently.
+    - Creating an access claim for the file.
+    - Priming the Work Package Service.
+    - Creating a Work Package.
+- Requesting the file download from the Download Controller Service.
+- Verifying the publication of the expected `download_served` event.
+- Requesting the envelope from the Download Controller Service.
+- Decrypting the downloaded output file and comparing the checksum.
+- Removing the test artifacts.
```
KerstenBreuer, line 41:
```suggestion
- Setting the initial state for the file to be downloaded:
    - Creating outbox and permanent storage bucket
    - place a file in the permanent storage bucket
    - Informing the Internal File Registry Service about the file stored permanently (event).
    - Creating an access claim for the file (REST call)
    - Priming the Work Package Service (event)
    - Creating a Work Package (REST call)
- Requesting the file download from the Download Controller Service (REST call via connector)
- Verifying the publication of the expected `download_served` event.
- Requesting the envelope from the Download Controller Service (REST call via connector).
- Decrypting the downloaded output file and comparing the checksum (using connector).
- Removing the test artifacts (via fixtures).
```
```suggestion
- Setting the initial state for the file to be downloaded:
    - Informing the Internal File Registry Service about the file stored permanently.
    - Creating an access claim for the file.
    - Priming the Work Package Service.
    - Creating a Work Package.
- Requesting the file download from the Download Controller Service.
- Verifying the publication of the expected `download_served` event.
- Requesting the envelope from the Download Controller Service.
- Decrypting the downloaded output file and comparing the checksum.
- Removing the test artifacts.
```

35-beluga-whale/technical_specification.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+## Additional Implementation Details:
+
+Before executing the tests, following setup should be achieved
+- Set vault for user
+- Set auth key
+- Set signing key for Work Package Service
+- Create storage buckets
```
KerstenBreuer, line 50: The setup of the storage buckets can be the first step of the test.

35-beluga-whale/technical_specification.md
```diff
@@ -0,0 +1,56 @@
...
+- Verifying the publication of the expected `download_served` event.
+- Requesting the envelope from the Download Controller Service.
+- Decrypting the downloaded output file and comparing the checksum.
+- Removing the test artifacts.
+
+
+## Additional Implementation Details:
+
+Before executing the tests, following setup should be achieved
+- Set vault for user
```
KerstenBreuer, line 47: What do you mean with the vault for user?

35-beluga-whale/technical_specification.md
```diff
@@ -0,0 +1,56 @@
...
+- Set auth key
+- Set signing key for Work Package Service
```
KerstenBreuer, line 49: Maybe we can setup an auth adapter fixture, that can be used to produce access tokens that are valid for the configured signing key. What do you think?

35-beluga-whale/technical_specification.md
```diff
@@ -0,0 +1,56 @@
...
+- Create storage buckets
+
```
KerstenBreuer, line 51:
```suggestion
Cleanup after test:
- Extend existing fixtures to clean up the test state after executing:
  - for MongoDB, drop used database
  - for s3 delete used buckets
  - for kafka delete used topics

```

KerstenBreuer (changes requested):
Looks great. Here are a few comments.

35-beluga-whale/technical_specification.md
dontseyit: Okay, that's wrong because I thought the init script which sets the vault in the test-bed is user-specific. We may say "Set Vault AppRole for EKSS"

35-beluga-whale/technical_specification.md
dontseyit: You mean the Work Package Access Token right? Because work order tokens are created upon request from WPS.

KerstenBreuer (approved):

PR #85, opened 2023-07-14
Tokay Gecko specs

mephenor:
High level overview, please comment if this needs more details.
Also added some rudimentary tickets in Jira that will be filled with details on monday.

### DIFF
--- a/38-wip/technical_specification.md
+++ b/38-wip/technical_specification.md
@@ -1,0 +1,72 @@
+# Missing Glue Code for Ingress Inter-service Communication (Tokay Gecko)
+**Epic Type:** Implementation Epic
+
+## Scope
+
+This epic aims to fill in the missing parts in inter-service communication along the preliminary upload/ingress path provided for 1.0.
+
+### Outline:
+
+#### Metldata Service:
+
+- Implement publisher for deletion and population events
+- Needs to know about MASS and WPS DB and be able to query them
+- Invesigate: Communication with MASS DB might be possible through existing API
+
+#### MASS:
+
+- Add event subscriber for deletion and population events
+- Add functionality to populate entities
+- Add functionality to delete entities
+- embedded_dataset artifact as input in form of MASS models.Resource
+- Kafka Key Name: dataset_ebmedded_{id}
+
+#### WPS:
+
+- Question to solve: Which service produces DatasetOverview?
+- Add event subscriber config for deletion and population events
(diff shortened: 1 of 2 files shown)


38-wip/technical_specification.md
```diff
@@ -0,0 +1,72 @@
...
+## Scope
+
+This epic aims to fill in the missing parts in inter-service communication along the preliminary upload/ingress path provided for 1.0.
+
+### Outline:
+
+#### Metldata Service:
+
+- Implement publisher for deletion and population events
+- Needs to know about MASS and WPS DB and be able to query them
```
Cito, line 13: Looking into foreign databases violates the principles of a microsevice architecture, do we really want to do that?

Instead, the metldata service should use an internal API to request that. Or even better, the metldata service could have its own database of already announced datasets to determine for which datasets deletion events need to be issued.

Actually, the WPS itself also doesn't need to be informed about deletions. It only needs the dataset overview events to know which files belong to which dataset in order to create the work package access token.

However, the claims repository may need to be informed about deletion, so that controlled access grants can be removed.

lkuchenb: IIRC the bottom line of the video call we had on the subject was that for this temporary implementation we would accept to look "behind the curtain". But actually we have all information already at metldata service anyhow if we fire the events before clearing the database so I guess that would anyway be the easier solution 🤔

38-wip/technical_specification.md
```diff
@@ -0,0 +1,72 @@
...
+- Invesigate: Communication with MASS DB might be possible through existing API
```
TheByronHimes, line 14:
```suggestion
- Investigate: Communication with MASS DB might be possible through existing API
```

TheByronHimes (changes requested):
Don't forget to rename the folder from 38-wip to 38-tokay-gecko.

TheByronHimes (approved):

38-wip/technical_specification.md
mephenor: So, this needs further discussion?

mephenor: @Cito What's the URL of the claims repo and what changes would be needed?

TheByronHimes (approved):

TheByronHimes (dismissed):

TheByronHimes (approved):

PR #87, opened 2023-09-18
Red Kangaroo epic

ac-jorellanaf:


31-red-kangaroo/technical_specifications.md
```diff
@@ -0,0 +1,44 @@
...
+
+## User Journeys (Optional)
+
+This epic covers the following user journeys:
+
+- The user views a list of all datasets
+- The user filters the list of datasets based on facets provided by the API
+- The user can see the summary of the details of the datasets the user selects.
+- The user can see the full details of the datasets the user selects
+- The user views a global summary of the entire metadata database
```
KerstenBreuer, line 31: I would think only the first two points are required to get a good idea. However, would like to hear @Cito 's opinion on that?

31-red-kangaroo/technical_specifications.md
```diff
@@ -0,0 +1,44 @@
...
+## Additional Implementation Details:
+
+- Although self-evident, it is important to specify the Metadata Catalog(ue) functionality will be implemented in both Angular and Vue.
+- The implementation will use the Metadata Repository and Search services used for the Metadata Catalog(ue) to avoid:
+  - Having to use the newer, more complicated loading of test data for the Metldata and MASS services
+  - Or have to immediately set up response mocking to allow the use of test data without an actual backend connection
+
+## Human Resource/Time Estimation:
+
+Number of sprints required: 3-4
```
KerstenBreuer, line 42: I would restrict this to 2 sprints ergo 1 month. When cutting down on the User Journeys as proposed, do you think this is possible @ac-jorellanaf and @Cito?

KerstenBreuer (changes requested):
Looks good, just some minor comments. I also tagged @Cito for review.

31-red-kangaroo/technical_specifications.md
Cito: Yes, let's keep it simple. The main point is to exercise how data is fetched from the backend and shown. It suffices to do it once.

What could be added is the form for submitting an access requests, in order to see how form handling is supported by the frameworks.

31-red-kangaroo/technical_specifications.md
Cito: I think it's ok to give it a bit more time if really both alternative frameworks shall be examined, and the re-implementation shall be done in a way that makes use of the concepts and tools provided by these frameworks, which first must be learned and understood. Can we agree on 3 sprints?

31-red-kangaroo/technical_specifications.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Principal Components of Exploration:
+
+This epic seeks to explore the use of Angular and Vue for a refactoring of the Data Portal (party due to the deprecation of Create-React-App) by implementing some basic Metadata Catalog(ue) functionality on both frameworks and deciding on a potential framework for future refactoring.
+
+### Questions to Answer
+- Are any of the two options suitable for the development of the data portal?
+- If so, which framework is best suited for the development of the data portal?
+- Which framework's data handling system is best suited for the requirements of the data portal?
+- Are Vue's advantages (e.g. shallower learning curve, less restrictive framework) enough of a positive to counter the lack of previous experience with it in the team?
```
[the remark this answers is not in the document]
Cito: - Which framework ensures that the code is well-organized, extendible and maintainable in the long-run and easy for new developers to understand?

31-red-kangaroo/technical_specifications.md
KerstenBreuer: 3 sprints is fine

31-red-kangaroo/technical_specifications.md
ac-jorellanaf: Changed :)

31-red-kangaroo/technical_specifications.md
ac-jorellanaf: Changed. I never know if the sprint count includes holidays, or if it is a more abstract count of development effort needed. Hence why I said 4.

KerstenBreuer (approved):

PR #90, opened 2023-10-30
Eurasian Wolf Epic (GSI-434)

KerstenBreuer:
Custom Specification Separating Schema Validation and Schema Linkage


lkuchenb (approved):
LGTM.

I'm not sure about the requirement to model & validate cardinality, at least in the submission model that should be implicit. For derived models with back references it may be useful though if the scope of validation should be beyond human generated data at time of submission.

sbilge (approved):

PR #91, opened 2023-11-27
Marabou Stork Epic for Correlation IDs (GSI 492)

TheByronHimes:
This technical specification is for the "Marabou Stork" epic, which concerns the work required to enable the use of correlation IDs in our services.

### DIFF
--- a/43-marabou-stork/technical_specification.md
+++ b/43-marabou-stork/technical_specification.md
@@ -80,9 +80,12 @@ retrieve the correlation ID from the ContextVar without any extra work.
 A pair of repositories (which is to be determined) will be updated as part of this epic.
 The purpose of this is to verify the aforementioned changes and identify any problems
 before rolling out the changes to other services. This should include one repository that utilizes
-a FastAPI app and a Kafka producer (such as the `upload-controller-service`) and another that
-features a Kafka consumer to process events published by the former
-(e.g. `interrogation-room-service`).
+a FastAPI app and a Kafka producer, and another that features a Kafka consumer to process events published by the former.
+
+Examples of such pairings:
+- `metldata` and Metadata Artifact Search Service (`mass`)
+- File Ingest Service (`fis`) and the Internal File Registry Service (`ifrs`)
+- Access Request Service (`ars`) and the Notification Service (`ns`)
 
 ## Human Resource/Time Estimation:
 
--- a/43-marabou-stork/technical_specification.md
+++ b/43-marabou-stork/technical_specification.md
@@ -1,0 +1,56 @@
+# Correlation ID Exploration (Marabou Stork)
+**Epic Type:** Implementation Epic
+
+Epic planning and implementation follow the
+[Epic Planning and Marathon SOP](https://docs.ghga-dev.de/main/sops/sop001_epic_planning.html).
+
+**Attention: Please do not put any confidential content here.**
+
+## Scope
+### Outline:
+The aim of this epic is to implement the functionality required to use interservice correlation IDs for tracking associated requests.
+Please see the prototype [here](https://github.com/ghga-de/prototype-correlation-id).
+
+### Included/Required:
+- Generate Correlation ID in API Gateway
+- Correlation ID Utilities in Hexkit
+- Correlation ID in Kafka Event Header
+- Middleware for FastAPI Apps
+
+### Not Included:
+- Updating services to use correlation IDs
+
+
+## Additional Implementation Details:
+
+### Generate Correlation ID in API Gateway
+The API Gateway will need to be updated with functionality to ensure every request has a correlation ID in the header.
+
+### Correlation ID Utilities in Hexkit
+ContextVars can only be referenced via a single instance, so it is not possible to instantiate a correlation ID ContextVar from one library and set it via another library without importing it. Therefore, `hexkit` should be made a production dependency of `ghga-service-commons`, and correlation ID utilities should be defined within `hexkit`.
+
+The following are needed in `hexkit`:
+- A ContextVar to store the correlation ID.
+- An asynccontextmanager to set the correlation ID ContextVar, yield, and then reset the token.
+- A function to retrieve the value stored in the ContextVar and raises an error if none is found.
+- A function that validates a correlation ID value and raises an error if the value is empty or invalid.
+
+### Correlation ID in Kafka Event Header
+There needs to be a way to supply the correlation ID as a Kafka event header. If the correlation ID is not supplied, one should be generated, but only if a flag has been explicitly set. Otherwise, an error should be raised.
+Kafka events have a header that can be used to store ancillary information, but the current Kafka providers and protocols don't allow for access to the event headers.
+Since it will be passed as an event header detail, correlation ID should be added as a parameter to the following:
+- `_publish_validated()` and `publish()` methods of the `EventPublisherProtocol` class
+- `_publish_validated()` method of the `KafkaEventPublisher` class
+
+The `_consume_event()` method in the `KafkaEventSubscriber` class needs to be updated to extract the correlation ID from the header, validate it (raising an error if invalid/missing), and set the ContextVar before calling `self._translator.consume()`
+
+### Middleware for FastAPI Apps
+A new middleware function for FastAPI apps should be added to `ghga-service-commons` to validate correlation IDs upon receiving requests.
+All requests should have a correlation ID already set by the gateway, so no logic is needed for ID generation.
+If the correlation ID is missing or invalid, an error should be raised. After getting the validated correlation ID, the middlware should set the correlation ID ContextVar using the utility created in `hexkit`. Doing so will allow services to retrieve the correlation ID from the ContextVar without any extra work.
+
+## Human Resource/Time Estimation:
+
+Number of sprints required: 1
+
+Number of developers required: 2
--- a/43-marabou-stork/technical_specification.md
+++ b/43-marabou-stork/technical_specification.md
@@ -8,7 +8,7 @@ Epic planning and implementation follow the
 
 ## Scope
 ### Outline:
-The aim of this epic is to implement the functionality required to use interservice correlation IDs for tracking associated requests.
+The aim of this epic is to implement the functionality required to use interservice correlation IDs for tracking request flows.
 Please see the prototype [here](https://github.com/ghga-de/prototype-correlation-id).
 
 ### Included/Required:
@@ -28,13 +28,14 @@ Please see the prototype [here](https://github.com/ghga-de/prototype-correlation
 The API Gateway will need to be updated with functionality to ensure every request has a correlation ID in the header.
 
 ### Correlation ID Utilities in Hexkit
-ContextVars can only be referenced via a single instance, so it is not possible to instantiate a correlation ID ContextVar from one library and set it via another library without importing it. Therefore, `hexkit` should be made a production dependency of `ghga-service-commons`, and correlation ID utilities should be defined within `hexkit`.
+In the prototype, we decided to use a ContextVar to store correlation IDs. This ContextVar needs to be created in a top-level module in either `hexkit` or `ghga-service-commons`. If it would be created in `ghga-service-commons`, then `hexkit` would need to import from there and have `ghga-service-commons` as a dependency, which we want to avoid since it should be a universal framework (also, we already have an inverse dependency, so we would create a circular dependency). This means the ContextVar must be created in `hexkit`.
 
 The following are needed in `hexkit`:
+- A function to create new correlation IDs.
 - A ContextVar to store the correlation ID.
-- An asynccontextmanager to set the correlation ID ContextVar, yield, and then reset the token.
-- A function to retrieve the value stored in the ContextVar and raises an error if none is found.
-- A function that validates a correlation ID value and raises an error if the value is empty or invalid.
+An asynccontextmanager to set the correlation ID ContextVar to a given value, yield it, and then reset it to its previous state (via a token that was created when setting the value).
+- A function to retrieve the value stored in the ContextVar for the current context and raise an error if none is found.
+- A function that validates a correlation ID value, raising an error if the value is empty or does not have the expected format.
 
 ### Correlation ID in Kafka Event Header
 There needs to be a way to supply the correlation ID as a Kafka event header. If the correlation ID is not supplied, one should be generated, but only if a flag has been explicitly set. Otherwise, an error should be raised.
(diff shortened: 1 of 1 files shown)


43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+ContextVars can only be referenced via a single instance, so it is not possible to instantiate a correlation ID ContextVar from one library and set it via another library without importing it. Therefore, `hexkit` should be made a production dependency of `ghga-service-commons`, and correlation ID utilities should be defined within `hexkit`.
+
+The following are needed in `hexkit`:
+- A ContextVar to store the correlation ID.
+- An asynccontextmanager to set the correlation ID ContextVar, yield, and then reset the token.
+- A function to retrieve the value stored in the ContextVar and raises an error if none is found.
+- A function that validates a correlation ID value and raises an error if the value is empty or invalid.
+
+### Correlation ID in Kafka Event Header
+There needs to be a way to supply the correlation ID as a Kafka event header. If the correlation ID is not supplied, one should be generated, but only if a flag has been explicitly set. Otherwise, an error should be raised.
```
Cito, line 40: Do we really need to supply the correlation ID to the publish method? I thought it would use the value from the context var, or create one (depending on a publisher config flag)? In both cases, no input would be needed?

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+## Additional Implementation Details:
+
+### Generate Correlation ID in API Gateway
+The API Gateway will need to be updated with functionality to ensure every request has a correlation ID in the header.
+
+### Correlation ID Utilities in Hexkit
+ContextVars can only be referenced via a single instance, so it is not possible to instantiate a correlation ID ContextVar from one library and set it via another library without importing it. Therefore, `hexkit` should be made a production dependency of `ghga-service-commons`, and correlation ID utilities should be defined within `hexkit`.
+
+The following are needed in `hexkit`:
+- A ContextVar to store the correlation ID.
```
Cito, line 34: - A function to create new correlation IDs.

Cito: This function could be used in the event publisher and the api gateway.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+- A function that validates a correlation ID value and raises an error if the value is empty or invalid.
+
+### Correlation ID in Kafka Event Header
+There needs to be a way to supply the correlation ID as a Kafka event header. If the correlation ID is not supplied, one should be generated, but only if a flag has been explicitly set. Otherwise, an error should be raised.
+Kafka events have a header that can be used to store ancillary information, but the current Kafka providers and protocols don't allow for access to the event headers.
+Since it will be passed as an event header detail, correlation ID should be added as a parameter to the following:
+- `_publish_validated()` and `publish()` methods of the `EventPublisherProtocol` class
+- `_publish_validated()` method of the `KafkaEventPublisher` class
+
+The `_consume_event()` method in the `KafkaEventSubscriber` class needs to be updated to extract the correlation ID from the header, validate it (raising an error if invalid/missing), and set the ContextVar before calling `self._translator.consume()`
```
Cito, line 46:
```suggestion
The `_consume_event()` method in the `KafkaEventSubscriber` class needs to be updated to extract the correlation ID from the header, validate it (raising an error if invalid/missing), and set the ContextVar during the call of `self._translator.consume()`
```

Cito: Should be reset after the call.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Middleware for FastAPI Apps
+A new middleware function for FastAPI apps should be added to `ghga-service-commons` to validate correlation IDs upon receiving requests.
+All requests should have a correlation ID already set by the gateway, so no logic is needed for ID generation.
```
Cito, line 50: We probably need a flag in the ApiConfigBase to control that behavior, since it should be also possible to test services without API gateway. If that flag is set, the ID would be created (using the function from hexkit), otherwise an error would be raised.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+If the correlation ID is missing or invalid, an error should be raised. After getting the validated correlation ID, the middlware should set the correlation ID ContextVar using the utility created in `hexkit`. Doing so will allow services to retrieve the correlation ID from the ContextVar without any extra work.
```
Cito, line 51:
```suggestion
If the correlation ID is missing or invalid, an error should be raised. After getting the validated correlation ID, the middelware should set the correlation ID ContextVar using the utility created in `hexkit`. Doing so will allow services to retrieve the correlation ID from the ContextVar without any extra work.
```

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Pilot Repository
+One repository (which is to be determined) will be updated as part of this epic. The purpose of this is to verify the aforementioned changes and identify any problems before rolling out the changes to other services. This should be a repository that utilizes a FastAPI app and both a Kafka producer and consumer.
```
Cito, line 54: Maybe it's easier to add it on one service with a REST API and a producer, and another service with a corresponding consumer. Like for instance, FIS and IFRS. We can then test the complete flow.

Cito (changes requested):
Thanks, sound like a good plan. Added comments regarding some details only.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,60 @@
...
+**Epic Type:** Implementation Epic
+
+Epic planning and implementation follow the
+[Epic Planning and Marathon SOP](https://docs.ghga-dev.de/main/sops/sop001_epic_planning.html).
+
+**Attention: Please do not put any confidential content here.**
+
+## Scope
+### Outline:
+The aim of this epic is to implement the functionality required to use interservice correlation IDs for tracking associated requests.
```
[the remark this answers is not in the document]
Cito: Since we are also tracking events, not only requests.

43-marabou-stork/technical_specification.md
Cito: Sorry for being confusing. With "during" I did not mean "inside" but exactly what you're doing here. I.e. not only setting it before the call, but also cleaning it up afterwards.

43-marabou-stork/technical_specification.md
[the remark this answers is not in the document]
TheByronHimes: Changed

43-marabou-stork/technical_specification.md
TheByronHimes: No, you're right it doesn't. I'm not sure why I got hung up on that. It should do like you said and grab the value from the ContextVar, then pass that with the _publish_validated() call.

43-marabou-stork/technical_specification.md
TheByronHimes: Had planned this but forgot to include it here. I've added it now, thanks.

43-marabou-stork/technical_specification.md
TheByronHimes: Good idea. Please see updates

43-marabou-stork/technical_specification.md
TheByronHimes: I like this idea. I've updated the spec accordingly, please review.

43-marabou-stork/technical_specification.md
TheByronHimes: No worries, thank you.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,91 @@
...
+or `ghga-service-commons`. If it would be created in `ghga-service-commons`, then `hexkit`
+would need to import from there and have `ghga-service-commons` as a dependency, which we
+want to avoid since it should be a universal framework (also, we already have an inverse
+dependency, so we would create a circular dependency). This means the ContextVar must be
+created in `hexkit`, which should be added as a dependency in `ghga-service-commons[api]`.
+
+The following are needed in `hexkit`:
+- A function to create new correlation IDs.
+- A ContextVar to store the correlation ID.
+An asynccontextmanager to set the correlation ID ContextVar to a given value,
```
Cito, line 44:
```suggestion
- An asynccontextmanager to set the correlation ID ContextVar to a given value,
```

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,91 @@
...
+There needs to be a way to supply the correlation ID as a Kafka event header.
+If the correlation ID is not supplied, one should be generated,
+but only if a flag has been explicitly set. Otherwise, an error should be raised.
+To that end, `hexkit` needs to be updated so the correlation ID is retrieved
+from the ContextVar for the current context and added as a header inside of the
+`publish()` method of the `KafkaEventPublisher` class. That method should also be
+updated to have a boolean parameter to act as the aforementioned flag.
+The correlation ID should be added as a parameter to the `_publish_validated()`
+method of both the `EventPublisherProtocol` and `KafkaEventPublisher` classes.
+
```
Cito, line 62: Alternatively, the flag could be also provided as a config setting of the publisher (can be decided later).

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,91 @@
...
+invalid/missing), and set the ContextVar before calling `self._translator.consume()`
+
+### Middleware for FastAPI Apps
+A new middleware function for FastAPI apps should be added to `ghga-service-commons`
+to validate correlation IDs upon receiving requests. All requests should have a
+correlation ID already set by the gateway, so no logic is needed for ID generation.
+However, `ApiConfigBase` should get a new parameter to instruct the middleware to
+generate a new ID (using the function from `hexkit`) rather than raising an error.
+This way, testing can be carried out without an API gateway.
+If the correlation ID is missing or invalid, an error should be raised. After getting the
```
Cito, line 74: I think this sentence "If the correlation ID is missing" should be moved after "no logic is needed for ID generation".

Cito (approved):
Thanks for the changes. Made some more small suggestions.

mephenor (dismissed):

43-marabou-stork/technical_specification.md
KerstenBreuer: One thing to consider is that we might want to decide on a per publish-basis whether to generate a correlation ID or expect an existing one. The config would configure that globally for the entire publisher.

43-marabou-stork/technical_specification.md
KerstenBreuer: I think it was in your notes because we discussed both options for handling cases where we expect no corr ID to be set as context var (e.g. for cron jobs):
1. set a flag to not expect but autogenerate a corr id -> Advantage that we don't expose the generation logic to the core
2. directly provide a custom corr id to the publish method -> Advantage that we are even more flexible that option 1.

Since the advantage of option 1 is not really that big because it is really just `uuid4()` to generate  a new corr ID, I would slightly prefer option 2. However, I would leave the decision to you. Both is absolutely fine.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,91 @@
+# Correlation ID Exploration (Marabou Stork)
```
KerstenBreuer, line 1: This is an implementation epic right?
```suggestion
# Correlation ID Implementation (Marabou Stork)
```

KerstenBreuer (approved):

43-marabou-stork/technical_specification.md
KerstenBreuer: However, as mentioned below, testing could be a bit simplified when this would be solved as a config parameter.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,91 @@
...
+
+The following are needed in `hexkit`:
+- A function to create new correlation IDs.
+- A ContextVar to store the correlation ID.
+- An asynccontextmanager to set the correlation ID ContextVar to a given value,
+yield it, and then reset it to its previous state (via a token that was created
+when setting the value).
+- A function to retrieve the value stored in the ContextVar for the current context
+and raise an error if none is found.
+- A function that validates a correlation ID value, raising an error if the value is
```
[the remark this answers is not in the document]
TheByronHimes: It should be, yes. I hadn't included that but I will.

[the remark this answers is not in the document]
KerstenBreuer: No worries, I just asked to understand. Anyhow, these details are better discussed in a code PR.

43-marabou-stork/technical_specification.md
```diff
@@ -0,0 +1,91 @@
...
+This way, testing can be carried out without an API gateway.
+After getting the validated correlation ID, the middleware should set the correlation
+ID ContextVar using the utility created in `hexkit`. Doing so will allow services to
+retrieve the correlation ID from the ContextVar without any extra work.
+
+### Pilot Repositories
+A pair of repositories (which is to be determined) will be updated as part of this epic.
+The purpose of this is to verify the aforementioned changes and identify any problems
+before rolling out the changes to other services. This should include one repository that utilizes
+a FastAPI app and a Kafka producer (such as the `upload-controller-service`) and another that
```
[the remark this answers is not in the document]
Cito: It's also not part of the archive test bed yet, so we cannot test it there. Some other examples are metldata (upload) + mass (as consumer), fis (ingest) + ifrs (as consumer) or ars (create request) + ns (as consumer), which are all tested already.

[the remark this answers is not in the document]
TheByronHimes: Thanks @Cito. I'll modify it it list those as example combinations instead.

43-marabou-stork/technical_specification.md
TheByronHimes: @Cito @KerstenBreuer  As long as there aren't any events that would get published outside of a context ID, I think the config option might be the way to go. It also keeps all the cid changes limited to the providers instead of the protocols. If you agree, I'll change the spec.

KerstenBreuer: Sounds reasonable

KerstenBreuer (approved):

Cito (approved):

PR #93, opened 2023-12-08
Structured Logging Epic - Topi Antelope (GSI-527)

TheByronHimes:
Adds the technical specification for the structured logging (Topi Antelope) epic.

Please let me know your thoughts.

image

### DIFF
--- a/45-topi-antelope/technical_specification.md
+++ b/45-topi-antelope/technical_specification.md
@@ -30,7 +30,11 @@ with other logs.
 There needs to be a way to easily invoke a logger object with minimal boilerplate code.
 There should also be a dedicated logging configuration class that subclasses `BaseSettings`.
 
-The configuration should at least include the log level for the service.
+The configuration should at least include:
+    - The log level for the service (`log_level`),
+    - The service name (`service_name`)
+    - The service instance ID (`service_instance_id`)
+
 When employing loggers in a service, it should be as easy as calling a function with a logger
 name as an argument and then using the object to log as needed.
 The logs should be emitted in a JSON string format (one line). An example log message is as
@@ -45,13 +49,13 @@ follows (formatted for ease of reading):
 	"correlation_id": "d826361b-3734-4590-b3e8-3cbed68b9236",
 	"message": "The file with ID 123xyz is already in the inbox.",
     "details": {
-		"file_id": "123xyz"
+		"file_id": "123xyz" // example detail
 	}
 }
 ```
-Additional configurable properties could include a way to disable uvicorn integration
-(e.g. for easier reading during development), or a format string that can be used to override
-the JSON output format.
+Additional configurable properties could include:
+    - A flag to control uvicorn integration (e.g. for easier reading during development)
+    - A format string which, if used, will replace the JSON output format.
 
 To summarize, this task includes all items in `hexkit`:
 - Logging Configuration Class
--- a/45-topi-antelope/technical_specification.md
+++ b/45-topi-antelope/technical_specification.md
@@ -1,0 +1,78 @@
+# Structured Logging Tools (Topi Antelope)
+**Epic Type:** Implementation Epic
+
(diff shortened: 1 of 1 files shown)


45-topi-antelope/technical_specification.md
```diff
@@ -0,0 +1,83 @@
...
+- Updating services to use new logging tools.
+
+
+## Additional Implementation Details:
+
+### Logging Tools in `Hexkit`
+There needs to be a way to easily invoke a logger object with minimal boilerplate code.
+There should also be a dedicated logging configuration class that subclasses `BaseSettings`.
+
+The configuration should at least include the log level for the service.
```
KerstenBreuer, line 33: Maybe also include the service name and the service instance id in the config?

45-topi-antelope/technical_specification.md
```diff
@@ -0,0 +1,83 @@
...
+When employing loggers in a service, it should be as easy as calling a function with a logger
+name as an argument and then using the object to log as needed.
```
KerstenBreuer, line 35: I don't understand what is meant by this but happy to wait for the code :)

KerstenBreuer: However, I would avoid adding logging details to function signatures if this is what you have meant. Moreover, I currently don't see a need for multiple loggers, or do we need it?

45-topi-antelope/technical_specification.md
```diff
@@ -0,0 +1,83 @@
...
+	"correlation_id": "d826361b-3734-4590-b3e8-3cbed68b9236",
+	"message": "The file with ID 123xyz is already in the inbox.",
+    "details": {
+		"file_id": "123xyz"
+	}
+}
+```
+Additional configurable properties could include a way to disable uvicorn integration
+(e.g. for easier reading during development), or a format string that can be used to override
+the JSON output format.
```
KerstenBreuer, line 54: I like that. If it is simple to implement, we could also have a json_log_format boolean parameter in the config to switch between JSON and non-JSON (more readable) logging.

45-topi-antelope/technical_specification.md
```diff
@@ -0,0 +1,83 @@
...
+
+To summarize, this task includes all items in `hexkit`:
+- Logging Configuration Class
+- Logger factory
+- The subclass implementations of any support `logging` classes, such as
+Formatter and LoggerAdapter.
+
+### Consolidating Uvicorn Logging
+`ghga-service-commons` needs to be modified so uvicorn's logging configuration can be changed.
```
KerstenBreuer, line 63: We want to use the same logging config as in hexkit, right?
```suggestion
`ghga-service-commons` needs to be modified so uvicorn's logging can be configured using the same
config model used in hexkit.
```

KerstenBreuer (dismissed):
Looks good, just some minor questions.

45-topi-antelope/technical_specification.md
TheByronHimes: No, what I was referring to was instantiating a logger at the top of a module in order to use it throughout the module as needed. Usually you would call `logging.getLogger()` (maybe with `__name__` or something as an arg), then do the basic config before using it. What I meant here was that our logger(s) should be just as simple to use.

45-topi-antelope/technical_specification.md
TheByronHimes: 👍 I didn't think that would work since it's already in the KafkaConfig, but it seems to be okay. That will simplify a couple things.

45-topi-antelope/technical_specification.md
TheByronHimes: That sounds reasonable to me

Cito (approved):
Looks good. Just fixed some markdown issues caused by wrong indentations.

Cito (approved):

PR #96, opened 2024-01-26
Fix directory names and add table of contents

Cito:
- make the names of the subdirectories more consistent: replaced underscores with dashes since that was what most of them were using
- add script to print a table of contents
- add table of contents to the readme


create_toc.py
```diff
@@ -0,0 +1,50 @@
...
+                title = line[1:].strip()
+                break
+        else:
+            title = None
+        if not title:
+            raise RuntimeError(f"No title found in {spec_path}")
+        try:
+            desc, name = title.rsplit("(", 1)
+            desc = desc.rstrip()
+            name = name[:-1].strip().capitalize()
```
TheByronHimes, line 33:
```suggestion
            name = name[:-1].strip().title()
```
I think title case looks a little nicer with multi-word titles (Wood Ant vs Wood ant). It's a style choice so take it or leave it.

create_toc.py
```diff
@@ -0,0 +1,50 @@
...
+            num = int(epic_dir.split("-", 1)[0])
+            if not 0 <= num < 1000:
+                raise ValueError
+        except ValueError:
+            raise RuntimeError(f"Invalid directory name {epic_dir}")
+        spec_path = os.path.join(epic_dir, "technical_specification.md")
+        if not os.path.exists(spec_path):
+            raise RecursionError(f"No technical_specification.md found in {epic_dir}")
+        for line in open(spec_path):
+            if line.startswith("#"):
```
TheByronHimes, line 23:
```suggestion
            if line.lstrip().startswith("#"):
```
If you're bothering to strip the name below, then you might as well strip any spaces that wind up in front of the line so it doesn't try to read the scope header as the title.

TheByronHimes (changes requested):
One small change I think should be included and one suggestion. 
This seems like something that could be a pre-commit hook, too.

create_toc.py
Cito: Actually the line needs to start with a `#` but I get your point and fixed it in 6cbe3da.

Cito:
If the line numbering is non-continuous, then Markdown makes it continuous which is not what we want. Made it an unordered list in that case to preserve the original numbering.

Cito:
We can consider using a pre-commit hook and a script that auto-generates the readme after moving to the GitLab based documentation.

TheByronHimes (approved):

PR #99, opened 2024-02-13
Improve session management

Cito:
This PR revises the Sugar Ant epic spec, improving the session management:

- Refinement and explanation of the possible session states
- Explicit session creation via a new `/rpc/login` endpoint

The new endpoint will create a new session from a given OIDC access token or retrieve the existing backend session, so that the state from the backend can be transported to the frontend on a page refresh.

In principle, it would be possible to keep track of the state in memory and save it in sessionStorage on the client, but sessionStorage is unfortunately not replicated to different tabs. It would be probably possible to replicate the state in other tabs via an events-based mechanism, but this is difficult and error-prone and may not work equally well with all browsers. We do not want to use localStorage, because that would store the state permanently in the browser. Using an endpoint to fetch the current state from the backend seems to be the cleanest and most accurate solution.

This also allows to remove the endpoint for checking the existence of a TOTP token, since it can now be derived from the backend session state.


46-sugar-ant/technical_specification.md
```diff
@@ -258,49 +301,37 @@ The `AccessRequest` model must be extended to also include an optional `iva_id`
...
+If the session state is `registered`, the user will be shown the registered data (maye with an option to change them via re-registration) and informed that a second factor needs to be created. If the user confirms, the state progresses to `needs-totp-token`.
 
-Let's assume the user is in the `needs-totp-token` or `lost-totp-token` state. The frontend then uses the `POST /totp-token` endpoint to create a provisioning URI, whereby the `force` flag should be set if and only if the user is in the state `lost-totp-token`. The frontend then presents the returned URI in form of a QR code to the user and asks the user to scan the QR code using an authenticator app. It should also show a button or link to display the secret as text as fallback for manually entering the secret.
+The frontend then uses the `POST /totp-token` endpoint to create a provisioning UR, presents it in form of a QR code to the user and asks the user to scan the QR code using an authenticator app. It should also show a button or link to display the secret as text as fallback for manually entering the secret and as backup code.
```
ac-jorellanaf, line 326:
```suggestion
The frontend then uses the `POST /totp-token` endpoint to create a provisioning URL, presents it in form of a QR code to the user and asks the user to scan the QR code using an authenticator app. It should also show a button or link to display the secret as text as fallback for manually entering the secret and as backup code.
```

ac-jorellanaf (approved):

KerstenBreuer (approved):
Looks good. However, maybe localStorage would also be an option since many website support a continued login even if you have closed the browser. If I understand correctly this is also only about user metadata so nothing security relevant. However, I think this solution is also very good.

Cito:
It's true, the session data is not security critical, only the cookie is critical, which is tied to the browser session and not permanent. This means, when keeping the data in localStorage, after a restart, the browser would then still believe that it is logged on, but the cookie is already gone.

PR #101, opened 2024-02-27
Update tech spec of Sugar Ant (GSI-637)

Cito:
Some updates to the Sugar Ant tech spec:
- Slightly changed CSRF protection mechanism
- Details of Session model and X-Session
- Fixed backend flow image
- Mention another notification


46-sugar-ant/technical_specification.md
```diff
@@ -51,7 +51,7 @@ The Auth Adapter creates an auth session and tracks users as soon as they have l
...
+The Auth Adapter must also be extended with a mechanism that prevents "session riding" attacks using CSRF tokens. Thereby, the first request that creates the session responds also creates a CSRF token, which should be a unique and unpredictable string stored as part of the session. Contrary to the session cookie, the CSRF token must be made known to the frontend application, so that it can be passed in the request header as a CSRF token for each request to the backend. The Auth Adapter needs to compare the CSRF token with the unique string stored in the backend session.
```
ac-jorellanaf, line 54: Everything else is fine, but here maybe I would make sure to specify that the CSRF token should be re-generated after every request/page load?

ac-jorellanaf (approved):

46-sugar-ant/technical_specification.md
Cito: I think that is not needed and would also cause problems when using multiple tabs. See also here:
- https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html
- https://security.stackexchange.com/questions/22903/why-refresh-csrf-token-per-form-request

PR #105, opened 2024-03-11
Minor corrections in the Sugar Ant specs

Cito:


46-sugar-ant/technical_specification.md
```diff
@@ -412,7 +423,11 @@ The following flow diagram visualizes the login flow in the frontend.
 
 The following flow diagrams visualize the backend flows for the various routes that are handled by the Auth Adapter.
 
-Note that per the ExtAuth protocol, a response with a status code of "200 OK" means that the route is considered valid by the API gateway and forwarded to the corresponding micro service. Any other status code in the response causes the response to be directly passed back to the client.
+Note that per the [ExtAuth](https://www.getambassador.io/docs/edge-stack/latest/topics/running/services/ext-authz) protocol used by Envoy-based proxies like Emissary-ingress, a response with a status code of `200 OK` means that the route is considered valid by the API gateway and forwarded to the corresponding micro service. Any other status code in the response causes the response to be directly passed back to the client. We utilise this behavior to let the Auth Adapter carry out a dual role, by communicating directly with the client in order to establish user sessions and enroll TOTP, and also regulating access to the backend while exchanging authorizating headers.
```
TheByronHimes, line 426:
```suggestion
Note that per the [ExtAuth](https://www.getambassador.io/docs/edge-stack/latest/topics/running/services/ext-authz) protocol used by Envoy-based proxies like Emissary-ingress, a response with a status code of `200 OK` means that the route is considered valid by the API gateway and forwarded to the corresponding microservice. Any other status code in the response causes the response to be directly passed back to the client. We utilise this behavior to let the Auth Adapter carry out a dual role, by communicating directly with the client in order to establish user sessions and enroll TOTP, and also regulating access to the backend while exchanging authorization headers.
```

TheByronHimes (approved):
Let me know if you update this for the 403 responses like we discussed and want another quick review.

TheByronHimes (approved):

PR #106, opened 2024-03-19
details added to transpiler refactoring

sbilge:


49-jackal/technical_specification.md
```diff
@@ -151,7 +151,8 @@ III. Transpile schemapack-based models to a spreadsheet representation
 
 - Make transpiler produce datapack instead of original JSON format
 - Migrate ghga-transpiler to a more general configuration strategy that ideally lives longer than individual metadata models
-  - Refactor transpiler to read relevant information from hidden sheet in XLS rather than config maintained in transpiler code base
+  - Refactor the transpiler to read the infomation on i) the data types (e.g., list, str, int), ii) the value restriction (unrescticted, controlled vocabulary, value from another class) from a hidden sheet in the XLS file. This information will be presented in table form with boolean values.
```
lkuchenb, line 154:
```suggestion
  - Refactor the transpiler to read the infomation on i) the data types (e.g., list, str, int), ii) the value restriction (unrestricted, controlled vocabulary, value from another class) from a hidden sheet in the XLS file. This information will be presented in table form with boolean values.
```

lkuchenb (changes requested):
Thanks, but it would be great if we could really specify the format here (sheet name, column names, etc) to the necessary detail such that we can work in parallel on the implementations

lkuchenb (dismissed):
Format has in the meantime been agreed upon

lkuchenb (approved):

PR #108, opened 2024-03-25
Add transformation aggregation replacement logic

lkuchenb:
This sub-PR adds the migration of the current aggregate transformation logic.

### DIFF
--- a/50-dhole/technical_specification.md
+++ b/50-dhole/technical_specification.md
@@ -240,10 +240,134 @@ B. For merging relation properties:
   ```
 
 ### Refactor aggregate transformation:
-TBD.
+
+The former `aggregate` transformation shall be replaced by a mixture of existing
+and to be written transformations. The following new transformations shall be
+implemented:
+
+#### Transformation 1: Add Subschema
+
+A transformation that enables the insertion of a subschema into a new property
+of any object within the current content schema of a class, including an initial
+value. The configuration may look as follows:
+
+```yaml
+# Generic Example
+- class: MyClass
+  target_path: [path, of, properties]
+  schema:
+    type: "object"
+    additionalProperties: false
+  value: {}
+
+# Examples from current config
+- class: Dataset
+  target_path: [datasets]
+  schema:
+    type: object
+    additionalProperties: false
+    properties:
+      stats:
+        type: object
+        additionalProperties: false
+- class: Dataset
+  target_path: [studies_summary]
+  schema:
+    type: object
+    additionalProperties: false
+    properties:
+      stats:
+        type: object
+        additionalProperties: false
+```
+
+* The transformation shall check whether the specified `target_path` is valid, i.e. all but the last elements are objects and the property does not yet exist.
+
+#### Transformation 2: Count References
+
+* The transformation shall count how many target objects are referenced from each source object given the reference name.
+* The transformation shall validate whether the target is defined with multiplicity and fail otherwise
+
+Example config:
+
+```yaml
+- class: Dataset
+  target_path: [samples_summary, count]
+  source_path: [samples]
+```
+
+####  Transformation 3: Count content values
+
+* The transformation shall count the values encountered at a specified property in the content of an object.
+* The transformation shall validate that at least one of the traversed references is multi-valued by its cardinality or one of the traversed content elements is an array.
+* The path to the content property is specified as the list of references to traverse (`source_path.reference`) and the list of content objects to traverse (`source_path.content`).
+* The terminal element of the `source_path.content` configuration is the name of the property to be created to hold the resulting value. It is added to the schema by the transformation and has the following schema:
+  ```yaml
+  type: object
+  additionalProperties: true
+  ```
+  where each observed value will be mapped to an integer value representing the
+  number of times it was observed
+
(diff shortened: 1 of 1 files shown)


KerstenBreuer (commented):
I am sorry, but I cannot give good feedback because I struggle to understand the proposed transformations. I would be happy to go through them in a meeting. Otherwise I also trust you that they are what we need conceptually. However, we might still want to agree on common naming patterns. An whether transformation can be used to perform a batch of tasks (e.g. modify multiple classes) or shall only focus at one task at a time.

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,134 @@ B. For merging relation properties:
...
+- class: MyClass
+  target_path: [path, of, properties]
+  schema:
+    type: "object"
+    additionalProperties: false
+  value: {}
+
+# Examples from current config
+- class: Dataset
+  target_path: [datasets]
```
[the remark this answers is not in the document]
lkuchenb: Fixed, that was wrong indeed.

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,134 @@ B. For merging relation properties:
...
+#### Transformation 1: Add Subschema
+
+A transformation that enables the insertion of a subschema into a new property
+of any object within the current content schema of a class, including an initial
+value. The configuration may look as follows:
+
+```yaml
+# Generic Example
+- class: MyClass
+  target_path: [path, of, properties]
```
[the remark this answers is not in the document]
lkuchenb: Yes, data could look like that after the insertion of the subschema.

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,134 @@ B. For merging relation properties:
...
+```yaml
+- class: Dataset
+  target_path: [samples_summary, count]
+  source_path: [samples]
+```
+
+####  Transformation 3: Count content values
+
+* The transformation shall count the values encountered at a specified property in the content of an object.
+* The transformation shall validate that at least one of the traversed references is multi-valued by its cardinality or one of the traversed content elements is an array.
```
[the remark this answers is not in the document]
lkuchenb: Sorry for the wrong wording, references are relations, I will update the spec accordingly. In order to be able to count anything, the counting transformations shall validate that there is actually a case of multiplicity anywhere in the source path, either during relation traversal or during content traversal.

lkuchenb:
@KerstenBreuer Ready for another round I think

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,141 @@ B. For merging relation properties:
...
+##### Configuration
+
+```yaml
+# Generic Example, defaults shown for optional parameters
+- class_name: MyClass
+  target_content:
+    object_path: "" # Optional, default empty string refers to content root object
+    property_name: new_property
+    required: true # Optional
+  schema: # Optional
```
KerstenBreuer, line 263: Right?
```suggestion
  schema: # Optional, defaults to the following value
```

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,141 @@ B. For merging relation properties:
...
+
+A transformation that enables the insertion of a subschema into a new property
+of any object within the current content schema of a class, including an initial
+value. The schema will be modified by inserting the subschema at the specified path. The data will be modified by inserting the default value at the specified path. The user may control whether or not the newly added property is added to the parent property's `required` list.
+
+##### Configuration
+
+```yaml
+# Generic Example, defaults shown for optional parameters
+- class_name: MyClass
```
KerstenBreuer, line 258: Should we stay with arrays of instructions for our transformation steps? I am also happy to defer that decision to later.

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,141 @@ B. For merging relation properties:
...
+    property_name: datasets
+
+- class_name: Dataset
+  target_content:
+    property_name: "studies_summary"
+```
+
+#### Transformation 2: Count References
+
+* The transformation shall count how many target objects are referenced from each source object given the reference name.
```
KerstenBreuer, line 280:
```suggestion
* The transformation shall count how many target objects are referenced from each source object given the relation name.
```

50-dhole/technical_specification.md
```diff
@@ -240,10 +240,141 @@ B. For merging relation properties:
...
+* The transformation shall validate whether the target is defined with multiplicity and fail otherwise
+
+Example config:
+
+```yaml
+- class_name: Dataset
+  target_content:
+    object_path: samples_summary
+    property_name: count
+  source_relation_path: "Dataset(samples)>Sample"
```
KerstenBreuer, line 290: Great, actually I just thought that in deed passive relations (containing <) might also be interesting to count. This notation would enable it.

KerstenBreuer (approved):
Looks good, looking forward ot see it in action.

50-dhole/technical_specification.md
KerstenBreuer: As discussed, will keep the current representation.

PR #109, opened 2024-05-16
Update Sugar Ant spec

Cito:


TheByronHimes (approved):
There's a small typo on line 114 where rpc is written as rpc, but it looks good otherwise.

PR #110, opened 2024-05-21
Update tables and touch up paragraphs (GSI-758)

TheByronHimes:
The epic spec was fairly outdated, so I added some additions to keep it current. I will not continue to update this spec, but I thought it would be good to make a one-time update. I deleted the `Source Exists` column from the tables because they weren't useful.


mephenor (changes requested):
Phrasing talking about the event models should be slightly adjusted. the rest looks good.

mephenor (approved):

PR #111, opened 2024-06-03
Hero Shrew - Epic for Outbox Refactoring Implementation (GSI-777)

TheByronHimes:
This epic is for the implementation of the outbox pattern with the aim of backing up events

### DIFF
--- a/52-hero-shrew/technical_specification.md
+++ b/52-hero-shrew/technical_specification.md
@@ -17,7 +17,7 @@ republishing and reprocessing of these events should then result in the re-creat
 all transitive events, as far as idempotence allows.
 
 The outbox pattern may be implemented to make information about domain objects available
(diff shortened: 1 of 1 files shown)


Cito (dismissed):

mephenor (changes requested):
For the current upload path the File Ingest Service is also relevant, i.e. there should be an outbox publisher and corresponding outbox subscriber in the Internal File Registry for the `FileUploadValidationSuccess` event. This means the Interrogation Room also needs an Outbox Publisher for that event, until we completely move to the Connector based upload in the future.

Implementing the changes for the Upload Controller/Interrogation Room could be deferred until we actually need them, as they are currently unused, but that's up to you. If they are not implemented right now, this should be noted in the refactoring log.

mephenor (approved):

PR #113, opened 2024-06-05
Storage selection for DS Kit

mephenor:
This epic describes a way to move the storage alias selection for the preliminary upload path from the File Ingest Service to the Datasteward Kit.
If there's more that needs to be changed apart from where the alias is specified and how it is propagated and validated, let me now.


lkuchenb:
Thanks @mephenor!

Regarding the question how to provide the storage alias - the proposed solutions require that the user manually keeps track of where they uploaded the data in the previous step (`upload` or `batch-upload`). Wouldn't it make more sense to hook the mapping to storage aliases already into the upload process? i.e. the user configures the storage alias in their config instead of the URL when uploading (since you anyway intend to extend WKVS values we could publish a storage alias to URL map there, if we feel comfortable exposing the URLs there), the upload process then performs the upload and also annotates the storage alias in the upload metadata. Would that be an option?

mephenor:
I was thinking about keeping this information separate from the current output metadata, but from a usability context this probably makes more sense, I'll change the spec accordingly. Thanks @lkuchenb

55-madagascan-sunset-moth/technical_specification.md
```diff
@@ -0,0 +1,51 @@
...
+In addition, the Well Known Value Service needs to provide a new value at the existing
+
+- GET /values/{value_name}
+
+endpoint. 
+The proposed value name is `storage_aliases` and this endpoint should return a map of all configured storage aliases and their respective URLs.
+
+## Additional Implementation Details:
+
+The Datasteward Kit/File Ingest Service interaction for both non-secret ingest endpoints is changed slightly, by only including the storage alias in the payload sent and received. Theoretically this results in an additional overhead of calling the WKVS for each payload received, but as aliases should not be removed, results can simply be cached in memory and only re-requested when an alias is not found in the local cache.
```
lkuchenb, line 45: I think calling the WKVS once per DSKIT invocation is totally acceptable

lkuchenb (approved):

PR #124, opened 2024-09-17
Add IVA check for data stewards

Cito:
This PR makes the following amendments to the Sugar Ant (2FA and IVAs) epic:

- add notification when IVA is unverified
- add IVA and verification check for data steward role


mephenor (approved):

46-sugar-ant/technical_specification.md
```diff
@@ -251,6 +254,12 @@ The claims repository also has endpoints at
...
+
+This means that at least for one data steward, the verification of the corresponding IVA needs to be set manually in the database. This data steward can then also verify the IVAs of other data stewards via the API. The inter-service integration tests need to set the state of the first data steward via the "state management service" since this cannot be done with pure black-box testing via the API.
+
+Note that currently, this seeding also removes all existing data steward claims. If we decide to keep the data steward claims, we should still not set the IVA to `verified` state on creation, since data stewards who re-created their 2FA token and thereby unverified their IVA could simply remove their IVA and would be re-verified on the next start of the claims repository.
```
lkuchenb, line 261: Not sure why this is stressed here. Why would we ever bootstrap with "verified" state?

46-sugar-ant/technical_specification.md
```diff
@@ -279,6 +288,8 @@ There are only the following two exceptions where the token will be also added b
   - the user must be already registered
   - the `id` field of the auth context will contain the internal id of the user, and it must correspond to the `user_id` in the path
 
+The `role` field is set in the "token exchange" code that converts the OIDC token to an internal access token. This code checks whether the current user is active and has a valid claim granting a GHGA role. In addition to checking the validity of the claim, the code now also needs to check whether the claim has the `iva_id` field set which is associated with an existing IVA in the verified state.
```
lkuchenb, line 291: Wasn't this already implemented before the latest change? Is this just a clarification of the epic doc?

lkuchenb (approved):

46-sugar-ant/technical_specification.md
Cito: It would simplify the bootstrapping with the first data steward and as an alternative to SMS for testing. But yes, maybe it's better to remove that sentence completely.

46-sugar-ant/technical_specification.md
```diff
@@ -406,9 +417,9 @@ Each existing IVA in the state `unverified` should have a button "request verifi
 
 After clicking "request verification", the state of the IVA should be moved from `unverified` to `code_requested`, a confirmation email should be sent to the user and a notification email should be sent to a data steward.
 
-After clicking "verify", the verification code should be requested from the user via an input field, and the state of the IVA should be moved from `code-transmitted` to `verified`. If this does not succeed, a corresponding error message must be shown to the user. After three failed attempts or when the verification code expired, users should be informed that they need to re-request the verification because the verification code expired.
+After clicking "verify", the verification code should be requested from the user via an input field, and the state of the IVA should be moved from `code-transmitted` to `verified`. If this does not succeed, a corresponding error message must be shown to the user. After three failed attempts or when the verification code expired, users should be informed that they need to re-request the verification because the verification code is no longer valid.
```
TheByronHimes, line 420:
```suggestion
After clicking "verify", the verification code should be requested from the user via an input field, and the state of the IVA should be moved from `code-transmitted` to `verified`, but only if the verification succeeds. If this does not succeed, a corresponding error message must be shown to the user. After three failed attempts or when the verification code expired, users should be informed that they need to re-request the verification because the verification code is no longer valid.
```

TheByronHimes (approved):

46-sugar-ant/technical_specification.md
Cito: changed in 864061

46-sugar-ant/technical_specification.md
Cito: The first two sentences describe the situation before, the last one the change that needs to be implemented.

Until now role type claims did not have an associated IVA, only data access grant claims used it.

46-sugar-ant/technical_specification.md
```diff
@@ -187,6 +187,7 @@ Additionally, the following RPC-style endpoints will be added:
     - `204 No Content`: state changed to `unverified`
     - `401 Unauthorized`: auth error (not a data steward)
     - `404 Not Found`: the IVA has not been found
+  - *should also send a notification to the user*
```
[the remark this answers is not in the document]
Cito: In the auth-service, it had been already implemented. It just was not documented here.

46-sugar-ant/technical_specification.md
lkuchenb: Oh I see. Thanks for clarifying.

lkuchenb (approved):

PR #128, opened 2024-11-05
DLQ Service (GSI-1073)

TheByronHimes:
Contains the plan for building a DLQ service. This iteration has no web UI, only a REST API that returns JSON.
We will be able to see what the next events are in the DLQ topic for a given service, then resolve them one at a time.


Cito:
Would it make sense to add the original topic name as a kind of prefix to the key in the DLQ and later remove it again?

TheByronHimes:
> Would it make sense to add the original topic name as a kind of prefix to the key in the DLQ and later remove it again?

That would probably work fine if you're referring to the compacted topic key clashing.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,300 @@
...
+  - *Returns a list of all configured services*
+  - Auth Header: internal token
+  - Response Body: array of strings (service names)
+  - Response Status: `200 OK`
+- `GET /topics/{service}`
+  - *Returns a list of all topics configured for the chosen service*
+  - Auth Header: internal token
+  - Response Body: array of strings (topic names)
+  - Response Status: `200 OK`
+- `GET /{service}/{topic}`
```
Cito, line 47: Should we add query params here to allow pagination (limit and offset)?

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,300 @@
...
+  - Response Status:
+    - `200 OK`: The event test was successful
+    - `422 Unprocessable Entity`: The non-empty request body is not a valid event. 
+    - `401 Unauthorized`: auth error (not authenticated)
+- `POST /{service}/{topic}`
+  - *Processes the next event in the topic, optionally publishing the supplied event*
+  - Auth Header: internal token
+  - Request Body: empty
+  - Response Body: empty or JSON representation of corrected event
+    - Need to consider whether checks should be added for fields like correlation ID.
```
Cito, line 71: Shouldn't this go to the *request* body?

Cito (approved):
Looks good, see two small comments.

Maybe you should also clarify what you mean with "internal token" for auth.

61-matamata-turtle/technical_specification.md
TheByronHimes: Yes, that's a typo

61-matamata-turtle/technical_specification.md
TheByronHimes: I've made the change

Cito (dismissed):
Looks like you adapted the names to match those from MASS. Good idea.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,296 @@
...
+  the DLQ topic. This is crucial functionality that enables the user to decide how to
+  resolve the next event. The number of events previewed is capped by configuration.
+
+### DLQ Topic Arrangement
+
+For a given Kafka topic, there will be a separate DLQ topic *per service*.
+For example, the UCS, DCS, and IFRS subscribe to the topic containing File Deletion
+Requested events. They would publish failed File Deletion Requested events to their own
+DLQ topics, e.g. `file-deletions.ucs-dlq`, `file-deletions.dcs-dlq`,
+`file-deletions.ifrs-dlq`. Each service has only one retry topic, however:
```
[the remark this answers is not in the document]
TheByronHimes: Reworded for clarity -- let me know if you think it needs further work

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,296 @@
...
+
+![DLQ Topic Arrangement](./images/topic%20distribution.png)
+
+### Event Ordering:
+Dead letter queues inherently present a potential threat to system-wide event ordering.
+However, ordering events by keys, the idempotent design of our services, and having
+separate DLQ topics for each original topic *per service* should prevent most problems.
+Additionally, only one partition should be configured per DLQ topic as there will be
+only one consumer for each DLQ topic.
```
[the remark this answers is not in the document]
TheByronHimes: Yes it is. I reworded this part as a requirement to sound less like a suggestion

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,296 @@
...
+Requested events. They would publish failed File Deletion Requested events to their own
+DLQ topics, e.g. `file-deletions.ucs-dlq`, `file-deletions.dcs-dlq`,
+`file-deletions.ifrs-dlq`. Each service has only one retry topic, however:
+
+![DLQ Topic Arrangement](./images/topic%20distribution.png)
+
+### Event Ordering:
+Dead letter queues inherently present a potential threat to system-wide event ordering.
+However, ordering events by keys, the idempotent design of our services, and having
+separate DLQ topics for each original topic *per service* should prevent most problems.
```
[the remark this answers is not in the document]
TheByronHimes: Just unknown unknowns or flawed event design on our part, otherwise no.

61-matamata-turtle/technical_specification.md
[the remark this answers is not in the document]
TheByronHimes: This is now outdated as there is only one global dlq topic. One retry topic per service is the same though.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,296 @@
...
+require a small modification to `hexkit`.
+
+**Previewed Events**:  
+Events returned by the preview endpoint will match the format defined by the
+[`ExtractedEventInfo`](https://github.com/ghga-de/hexkit/blob/e3cad2d57212d1ba897e620501d5d2b120c70338/src/hexkit/providers/akafka/provider/eventsub.py#L63) dataclass.
+
+**Other**:  
+The event ID header (consisting of the original topic, partition, and offset) will be
+featured in both inbound and outbound events.
+
```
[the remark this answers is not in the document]
TheByronHimes: Events published to the DLQ topic will have the original topic header

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+8. The DLQ Service returns the published event information to the client.
+9. The service consumes the previously-DLQ'd event from its retry topic.
+
+### DLQ Topic Arrangement
+There is one DLQ topic. The DLQ topic name is set in configuration for some degree of
+flexibility should it be needed, but all services should use the same configured value.
+The default is `dlq`.
+
+### Persisting DLQ Events
+The DLQ Service will continually listen to the DLQ topic. When it gets an event,
```
Cito, line 123:
```suggestion
The DLQ Service will continually consume from the DLQ topic. When it gets an event,
```

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+publisher, which is the only downside of doing it this way instead of generating a UUID
+right before publishing the event.
+
+The event ID is always created, but only propagated if it enters the DLQ cycle.
+This differs from the correlation ID, which is propagated across services and persisted
+in the database in the case of outbox events.
+
+To illustrate:
+
+![Event ID usage](./images/event%20ID.png)
```
Cito, line 153: Would be good to have little numbers in the diagram like in the other one.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+E.g.: `dcs,file-registrations,0,123` would indicate an event from the "file-registrations"
+topic, partition 0, offset 123. The information is only seen by the consumer, not by the
+publisher, which is the only downside of doing it this way instead of generating a UUID
+right before publishing the event.
+
+The event ID is always created, but only propagated if it enters the DLQ cycle.
+This differs from the correlation ID, which is propagated across services and persisted
+in the database in the case of outbox events.
+
+To illustrate:
```
Cito, line 151: Mention here that "nos" is our example service and "users" the example topic, where another services publishes and event to users, and nos tries to consume it.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+as long as events are designed to use the correct keys and topics in the first place.
+
+### Event identification:
+Right now, events can be identified through a combination of correlation ID + event
+type and topic. It would be easier if there were a single field to associate a given
+event in a DLQ topic with its source event in the original topic. The combination of
+service name, topic name, partition, and offset uniquely identify an event, so we
+can concatenate the values in a single string and pass it along as an event header.
+
+E.g.: `dcs,file-registrations,0,123` would indicate an event from the "file-registrations"
```
Cito, line 142: I wonder whether we should use a different separator that does not need to be encoded when passed in a URL, so the URL is more readable. We cannot use dot or hyphen since they can occur in topic names. Underscores should be possible according to our naming conventions, but we should not rely on that. Maybe a tilde `~`?

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+To illustrate:
+
+![Event ID usage](./images/event%20ID.png)
+1. An event is published to the "users" topic, where it is stored at partition 0, offset 17.
+2. The `hexkit` Kafka provider used by the `NOS` gets the message from `aiokafka`.
+3. The Kafka provider hands the transformed event over to the service-defined
+   translator, which performs some business logic driven by the event.
+4. The translator encounters an error which it cannot mitigate, and all retries fail.
+5. The Kafka provider logs the error, including the event ID (`"nos,users,0,17"`).
+6. The Kafka provider publishes the event with the event ID header to the DLQ topic.
```
Cito, line 160: Why is the DLQ in the diagram labelled "users.nos-dlq"? I thought it was global and just `dlq`?

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+### Event processing:
+When we resolve the next event in a given DLQ topic, it can go one of two ways:
+- We can discard the event, meaning it is not republished to a retry queue and is
+  effectively ignored.
+- We can process the event, where the event is published to the retry topic or
+  discarded (e.g. if the service info is missing). The processing can also be bypassed
+  by directly supplying a JSON event to the API, which publishes the event outright
+  as long as the correlation ID matches.
+
+Another useful piece of information for the DLQ service is the exception information.
```
Cito, line 183: In principle, the exception information should be available in the log of the service and we should be able to find it using the timestamp and event ID. So this would be just for convenience, so that it can be shown in the DLQ service client?

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+republishing the event to a retry topic.
+
+This [Java DLQ implementation](https://medium.com/nerd-for-tech/-to-re-queue-apache-kafka-dlq-messages-95941525ca77)
+uses such headers (toward the bottom).
+
+### Processing Dry Run:
+It would be nice to be able to perform a dry-run of event processing as a sanity check.
+The test endpoint would perform any applicable processing on the next event and
+return the result instead of publishing it. If processing would result in the
+event being discarded instead of published, then it will return nothing.
```
Cito, line 195: Above you already mention this as a feature to be implemented instead of a nice to have. Maybe just remove this section.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+    "field1": ...,
+    "field2": ...,
+    ...
+  },
+  "key": ...,
+  "headers": {
+    "correlation_id": ...,
+    "exc_class": ...,
+    "exc_msg": ...,
+    "id": <topic - partition - offset>
```
Cito, line 215: the id would also contain the service name, right? would be good to show the exact format of the id here again.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+return the result instead of publishing it. If processing would result in the
+event being discarded instead of published, then it will return nothing.  
+
+
+### Previewing Events:
+Previewed events will be formatted with the `ExtractedEventInfo` class and returned as 
+JSON:
+```json
+{
+  "topic": ...,
```
Cito, line 203: could also have a unique id (the one used in the database anyway) that can be passed to the POST and DELETE endpoints instead of the event ID

Cito (changes requested):
Well explained. Made some suggestions below.

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+    - `401 Unauthorized`: auth error (not authenticated)
+- `POST /{service}/{topic}`
+  - *Processes the next event in the topic, optionally publishing the supplied event.*
+    *Returns the payload passed to the publisher.*
+  - Auth Header: internal token
+  - Query parameters:
+    - `dry_run` *(bool)*: if True, the endpoint will not actually publish the event to
+    the retry topic or delete it from the database. This is useful for verifying what
+    will get published before resolving an event.
+  - Request Body: empty OR JSON representation of corrected event
```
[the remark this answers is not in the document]
TheByronHimes: That's correct

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+    - `422 Unprocessable Entity`: The non-empty request body is not a valid event. 
+    - `401 Unauthorized`: auth error (not authenticated)
+- `DELETE /{service}/{topic}`
+  - *Directly discards the next event in the topic*
+  - Auth Header: internal token
+  - Response Body: empty
+  - Response Status:
+    - `204 No Content`: The event has been discarded
+    - `401 Unauthorized`: auth error (not authenticated)
+
```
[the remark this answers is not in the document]
TheByronHimes: That's fair. Maybe we should replace the composite event ID with a UUID4? Then we could generate that UUID upon publish, similar to the CID, and debug-log it for fuller traceability (that'd be a hexkit change).
Given that discarding an event means we're simply done with it forever, should we be allowed to discard events even if they're not the "next" event for a topic?

61-matamata-turtle/technical_specification.md
```diff
@@ -0,0 +1,247 @@
...
+All endpoints should require an internal auth token.
```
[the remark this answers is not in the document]
TheByronHimes: Data stewards should never touch the DLQ. It should only be people from the architecture team. But I think our data stewards have no interest in interacting with the DLQ anyway, so modifying the auth token, commons, and auth service is not something we need to do on first implementation IMO

61-matamata-turtle/technical_specification.md
TheByronHimes: Like I said in my response to your comment on the DELETE endpoint, maybe this is enough pressure to simply switch to a UUID.

61-matamata-turtle/technical_specification.md
TheByronHimes: I just missed it

61-matamata-turtle/technical_specification.md
TheByronHimes: Yeah, it doesn't cost anything and I think it would be nice to have

Cito: Could we provide the information about the service name, topic, partition and offset in a similar way if we use a random UUID to identify the event?

61-matamata-turtle/technical_specification.md
Cito: OTOH the composite event id was more readable and meaningful when debugging. But maybe we can provide the same information in other fields, similar to the exception?

61-matamata-turtle/technical_specification.md
TheByronHimes: Sure. They could be separate fields:
```
{
    "topic": "dlq",
    "type_": "file_registered",
    "payload": {},
    "key": "",
    "timestamp": 123412341234,
    "headers": {
        "correlation_id": "f7b5b22b-ed2a-4603-9466-2e3ae4178e42",
        "event_id": "2be69da8-84c1-453f-9224-10f7fdeb9328" (or "dcs,file-registrations,0,123"),
        "service": "dcs",
        "original_topic": "file-registrations",
        "partition":0,
        "offset": 123,
        "exc_cls": "RuntimeError",
        "exc_msg": "useful error message"
    }
}
```

Alternatively, all those "helper" datum could go into a subdict called something like `_dev_info`, where the underscore helps to...underscore... that it's not meant to be used programmatically (just an idea, maybe it doesn't make sense).
```
{
    "topic": "dlq",
    "type_": "file_registered",
    "payload": {},
    "key": "",
    "timestamp": 123412341234,
    "headers": {
        "correlation_id": "f7b5b22b-ed2a-4603-9466-2e3ae4178e42",
        "event_id": "2be69da8-84c1-453f-9224-10f7fdeb9328" (or "dcs,file-registrations,0,123"),
        "service": "dcs",
        "original_topic": "file-registrations",
        "_dev_info": {
            "partition":0,
            "offset": 123,
            "exc_cls": "RuntimeError",
            "exc_msg": "useful error message"
        }
    }
}
```

61-matamata-turtle/technical_specification.md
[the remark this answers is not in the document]
TheByronHimes: Another thought: Since changing the event_id would technically be a breaking change, would it be acceptable to provide the 4 components as query or path parameters? 
E.g. `DELETE /{service}/{topic}?partition=0&offset=123` or  `DELETE /{service}/{topic}/0/123`?

61-matamata-turtle/technical_specification.md
TheByronHimes: This is also a possibility, and would keep the change within the DLQS

61-matamata-turtle/technical_specification.md
TheByronHimes: Per your idea below, why not leave that field like it is and include a `dlq_id` in previewed events?

61-matamata-turtle/technical_specification.md
Cito: Or maybe put `dlq_info` on the top level (flat is better than nested)?

61-matamata-turtle/technical_specification.md
Cito: If we don't use it as an id, then I would prefer separate fields for readability and parseability.

61-matamata-turtle/technical_specification.md
Cito: And another field `dlq_id` that would contain the UUID for the event?

61-matamata-turtle/technical_specification.md
TheByronHimes: Done

61-matamata-turtle/technical_specification.md
TheByronHimes: Done

61-matamata-turtle/technical_specification.md
TheByronHimes: The revised approach inside the DLQS will put all the dlq-specific info (partition, offset, exc info, etc.) into a top-level `dlq_info` field, and generate a DLQ-specific UUID for reference to put in a top-level `dlq_id` field.

61-matamata-turtle/technical_specification.md
TheByronHimes: For now, the event ID will stay like it is, but the distinct elements within it will be stored in separate fields when the DLQS stores the DLQ event in the DB.

61-matamata-turtle/technical_specification.md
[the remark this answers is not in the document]
Cito: Yes, that would work as well.

lkuchenb (approved):

PR #129, opened 2024-11-19
DB Versioning (GSI-1128)

TheByronHimes:
This epic spec deals with implementing database migrations.


62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,158 @@
...
+then we need to move to version 6. Version 6 is not treated as some special 'undo'
+version increment; the migration logic merely happens to move the data to the same
+structure contained in version 4.
+
+### Migration Structure
+
+![Migration structure](./images/db%20migrations%20white%20bg.png)
+
+In the image above, the green section indicates the top-level migration logic, most of
+which can be implemented in a library like `hexkit`.
```
Cito, line 114: Where it says "get docs" did you mean "get collections"?

The diagram should also show what happens if there is an error because otherwise it's not evident why we need temporary collections.

Cito: I like the idea of having a framework to handle the most frequent case of updating every document in one or more collections to a new schema version. But the framework should also provide a hook to execute more complex custom migrations, like merging two collections into one.

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Errors During Migration
+
+If an error prevents a migration from finishing, then we should discard the processed
+entries, unset the lock document, and log the error. The cleanup is straightforward if
+migrated documents are stored in a temporary collection that can be dropped, rather than
+modifying the original collection directly (in which case we would need to reverse
+changes, which could be difficult). It's important that we test migrations thoroughly
+to avoid extended downtime from unexpected errors.
+
```
Cito, line 129: If a service has multiple collections, we need to keep all temporary collections until the whole migration is finished and then replace all the collections.

But even that may not be sufficient. What happens if in a multi-step-migration, the step at the end fails?

Maybe we should just somehow clone or dump the database (collections) as a backup before running the complete multi-step migration and if any step fails, restore the backup?

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,158 @@
...
+### Testing Migrations
+
+Abstracted logic should be tested wherever it lives, like `hexkit`.
+Tests should cover the following, but the list is not exhaustive:
+- The locking mechanism
+- Error handling
+- Logging
+- Database version detection
+- Selecting the right migration to start with
+- What to do when no matching migration exists
```
Cito, line 139: The abstracted logic would work like a framework, i.e. the service specific code is called by the general migration code in the framework.

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,158 @@
...
+
+When we test individual migration code, like the code that will update collection A for
+the migration from database version X to Y, we should use some mock data that represents
+documents in the database.
+Tests should at least verify that the migration code applies the right changes.
```
Cito, line 144: Good point, to detect problems early on and not cause downtime later.

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,158 @@
...
+
+After preventing simultaneous migrations, we need to optimize how other instances
+operate while a migration is underway and consider how to handle read and 
+write requests. One way would be to accept some downtime and
+block service instances until the migration is complete. This might be a
+sufficient initial solution because database sizes are small enough that migrations will
+be completed quickly. If we have to eliminate downtime, we could perform shadow
+migrations and write the results to temporary collections while the old service
+version continues to handle requests. When the migration is complete, the old service
+can be taken offline and the collections swapped out. That's a little more complex.
```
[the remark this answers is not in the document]
Cito: Another way to reduce downtime would be using parallelization. If there are multiple tables, they could be processed in parallel. Each table could be also partitioned using the primary key and then each partition migrated in a parallel process. (E.g. using a hash of the primary key module number of partitions.)

Cito (changes requested):
Thanks for working on this, made some suggestions.

62-california-condor/technical_specification.md
TheByronHimes: I'm in agreement with the backup strategy, I think this is also common

62-california-condor/technical_specification.md
TheByronHimes: Yes exactly

62-california-condor/technical_specification.md
TheByronHimes: In the case of a failure at the end of a multi-step-migration, though, it depends on where the error is occurring. If we maintain all the data in temp collections and do the replacements all at once at the end, then errors are only a real problem if they occur somewhere during the replacement stage. We can do those inside a transaction, however, in which case we only lose some time. Then, we can either try to directly perform the collection replacement again or drop the temp tables and start the migration process anew.

62-california-condor/technical_specification.md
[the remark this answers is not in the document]
TheByronHimes: I totally agree on the stop/deploy approach. I'm not advocating for a complex be-everything solution here, only trying to anticipate the "what if we need to minimize downtime?" questions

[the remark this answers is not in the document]
TheByronHimes: I've updated this section

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,197 @@
...
+need to automate or enforce iteration through collections. In fact, that would hinder
+our ability to write migration logic involving multiple collections at once.
+
+### Errors During a Migration
+
+We should use MongoDB's support for distributed transactions (a.k.a. multi-document
+transactions) to automatically roll back failed migrations. Here's how we can do this 
+in a database containing two collections named `users` and `orders`:
+1. Drop `tmp_users` and `tmp_orders` if they exist.
+2. Create `tmp_users` and `tmp_orders` and apply the indexes from the original tables.
```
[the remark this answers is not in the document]
TheByronHimes: Updated!

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,197 @@
...
+3. Read data from `users` and write new data to `tmp_users`.
+4. Read data from `orders` and write new data to `tmp_orders`.
+5. If no errors, start a transaction.
+   1. Drop `users` and `orders`.
+   2. Rename `tmp_users` to `users`.
+   3. Rename `tmp_orders` to `orders`.
```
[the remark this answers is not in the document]
TheByronHimes: D'oh, you're right. I read that doc, but apparently not closely enough. I came to the same conclusion as your second comment and think it will work just as well.

62-california-condor/technical_specification.md
```diff
@@ -0,0 +1,192 @@
...
+5. Rename `users` to `old_users` and `orders` to `old_orders`.
+6. Rename `tmp_users` to `users` and `tmp_orders` to `orders`.
+7. Apply the indexes from `old_users` and `old_orders` to `users` and `orders`.
+8. Do some verification (count docs, examine content, double check indexes, etc.).
+9. Drop `old_users` and `old_orders`.
+
+If an error occurs at any point, drop all `tmp_` tables from this migration, log the
+error, unset the lock document, make sure the old tables have the original names, and
+exit.
+
```
Cito, line 158: Regarding prefixes, maybe `tmp_new_` instead of `tmp_` and `tmp_old_` instead of `_old` ?

62-california-condor/images/migration flowchart.png
Cito, line 1: There should be also an arrow (maybe dotted red) from "Run Migrations" to "Raise Error" indicating that "Run Migrations" could throw an error and that the service should not start in that case.

Cito (approved):
Looks good, just two minor suggestions.

Cito (approved):

PR #130, opened 2024-12-10
Download path response caching

mephenor:
The documentation for this epic provides a high level overview over the following points:
- Challenges and open questions wrt how caching behaviour should be implemented
- Endpoints that should provide cacheable responses
- A simplified overview of the current request cascade when using the GHGA Connector


Cito (approved):
Thanks Thomas. Very good summary of what I had in mind for this epic.

PR #136, opened 2025-02-17
Oryx/Kafka config standardization epic (GSI-1386)

TheByronHimes:
Discusses a way to standardize Kafka event topic/type configurations. 

Question: would it make sense to drop "Events" from the config class titles for brevity?

### DIFF
--- a/67-oryx/technical_specification.md
+++ b/67-oryx/technical_specification.md
@@ -71,15 +71,6 @@ concept. The result is a "shoehorned" process that actually needs a dedicated
 mechanism in `hexkit`. That mechanism, a class that mimics some of the outbox
 DAO's functionality, can be created in another epic.
 
-For this epic, **if** a config class in the chart below is *not*
-marked with "*(stateful)*" but the current corresponding event types *are* dictated
-by `hexkit`'s outbox pattern event types, **then**:
-1. The event type's configured value, after using the standardized classes, should be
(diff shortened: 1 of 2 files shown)


67-oryx/technical_specification.md
```diff
@@ -0,0 +1,386 @@
...
+
+For this epic, **if** a config class in the chart below is *not*
+marked with "*(stateful)*" but the current corresponding event types *are* dictated
+by `hexkit`'s outbox pattern event types, **then**:
+1. The event type's configured value, after using the standardized classes, should be
+   temporarily set to `upserted` for continuity.
+   - The value can be set to something natural once the new persistent publisher class
+     is implemented.
+2. These are examples of places where the future stateless outbox DAO would be used.
+   - Not exhaustive -- most non-outbox events would be eligible
```
lkuchenb, line 82: I would avoid introducing the dummy config parameter here and only add it once we actually refactor that

lkuchenb (changes requested):
Thank's, just minor remarks!

mephenor (dismissed):

67-oryx/technical_specification.md
TheByronHimes: That advice has been incorporated.

lkuchenb (approved):
LGTM!

PR #138, opened 2025-02-25
Persistent Kafka Publisher (Gemsbok) Epic - (GSI-1391)

TheByronHimes:
Epic spec for adding an event publisher that incorporates a DAO to save events in the database as they are published, without trying to force the functionality through the `MongoKafkaDaoPublisher`.


lkuchenb (approved):
LGTM. I'm a bit concerned regarding the ordering by timestamp since that's not as unambiguous as real serial number, however, it's unclear to me in how far that can be easily achieved with mongo. Generally speaking I don't think we need to repub in the original order since there will be only one document per ID. So we might think of dropping that part alltogether 🤔

PR #140, opened 2025-03-05
OpenTelemetry + Jaeger implementation

mephenor:


69-common-kingslayer/technical_specification.md
```diff
@@ -0,0 +1,51 @@
+# OpenTelemetry in file services (Common Kingslayer)
```
Cito, line 1: Alternative title: "Distributive tracing with OpenTelemetry and Jaeger"

You can narrow the implementation scope to file services, but I think the main difficulty is to create spans that run across services, because that needs context propagation. Unless you actually want to keep this epic small and support only intra-service spans.

69-common-kingslayer/technical_specification.md
```diff
@@ -0,0 +1,51 @@
...
+
+Epic planning and implementation follow the
+[Epic Planning and Marathon SOP](https://docs.ghga-dev.de/main/sops/sop001_epic_planning.html).
+
+## Scope
+### Outline:
+
+Based on some previous exploration work, this epic aims to implement 1) OpenTelemetry instrumentation in the file services and 2) necessary OpenTelemetry functionality inside of hexkit and/or ghga-service-commons.
+In addition, a viable setup of Jaeger and its components needs to be devised and implemented.
+
```
Cito, line 12: Again, specify whether you want to support spans that involve multiple file services. If yes, it makes sense to also extend it to other services like the claims repo or even the api gateway or auth adapter.

The scope could be reduced to implement only one or two larger spans that contain both sync and async request flow (probably those that we have been struggling with in the last release), but with reusable code/providers in hexkit/service-commons.

69-common-kingslayer/technical_specification.md
```diff
@@ -0,0 +1,51 @@
...
+### Included/Required:
+
+While OpenTelemetry supports tracing, metrics and logging, this epic will focus on enabling collection of distributed tracing data and neglect the other two aspects.
+
+#### File Services:
+File services can export tracing data using manual or autoinstrumentation or a combination of both.
+Autoinstrumentation works by installing specific libraries that monkeypatch existing functionality to inject information that can be propagated across service boundaries.
+While these are quite quite useful, they should be vetted for what data is attached to the traces and manual instrumentation should be used instead, if sensitive information is included.
+If autoinstrumentation can be used in all/most cases, additional implementation work can be kept to a minimum.
```
Cito, line 21: Autoinstrumentation should be only used if it is well supported, documented and maintained, and manual instrumentation would be too complicated or time-consuming to implement. Otherwise we may run into problems when upgrading to a newer version of Python or the OpenTelemetry SDK.

69-common-kingslayer/technical_specification.md
```diff
@@ -0,0 +1,51 @@
...
+
+#### Hexkit/Service Commons:
+Hexkit needs some changes introducing logic around event subscribers to correctly propagate tracing context across service boundaries as the autoinstrumentation for Kafka does not open a new child span, ending the propagation at the event subscriber.
+
```
Cito, line 25: Is this not needed for synchronous requests?

69-common-kingslayer/technical_specification.md
Cito: Should this epic also quickly check alternatives to Jaeger, like Zipkin or Grafana Tempo?

69-common-kingslayer/technical_specification.md
Cito: If we add instrumentation to hexkit, then from the view of the services it would be also kind of auto-instrumentation. We should be very clear what we are talking about.

69-common-kingslayer/technical_specification.md
Cito: Would be good to specify which parts of the infrastructure should be covered by the instrumentation, e.g. S3, Mongo, sync and async calls. We want to see whether e.g. certain DB queries are very long.

Would be also good to show as an example how to trace certain code parts, e.h. calling a function that encodes or decodes a block of data.

Cito (changes requested):
Looks good, but I think the scope should be clearer or changed a bit.

Added some comments as food for thought, maybe we need to chat a bit to specify the exact scope.

69-common-kingslayer/technical_specification.md
mephenor: Not with autoinstrumentation enabled and does not need to be in hexkit.
Spans can be opened directly within in the FastAPI endpoint, but even that is not strictly necessary with autoinstrumentation enabled, span naming is just a bit more generic in that case.
If we do this manually, then we would want some middleware to handle context extraction and opening a parent span within the service.

I'll add a paragraph on that  to the docs.

69-common-kingslayer/technical_specification.md
mephenor: I'll take the second point into consideration, though autoinstrumentation should always mean via the libraries in this context. Might need to specify that in a sentence in the docs.

69-common-kingslayer/technical_specification.md
mephenor: I wouldn't include exploration of other backends for now.
Unsure if those are just plug and play replacements with some extra config or if there's more involved in getting those to run.
Current expertise from exploration is with Jaeger and if that's good enough for now, I would go with that. Either way it's not a critical part of the infrastructure and we can keep exploring the alternatives in the backlog for now.
To me that seems like a mini epic and if we should deem Jaeger unfit for our use case down the road, then we can reevaluate.

mephenor: Spans should cross service boundaries, yes.
I'll add details for other services. 
I would still implement that in the FSB first as there's existing code/config from exploration and then would move on to other services that would benefit from that afterwards.

mephenor: > Would be good to specify which parts of the infrastructure should be covered by the instrumentation, e.g. S3, Mongo, sync and async calls. We want to see whether e.g. certain DB queries are taking too long.
> 
> Would be also good to show as an example how to trace certain code parts, e.h. calling a function that encodes or decodes a block of data.

The first part I can extract from the exploration requirements.

What do you mean by the second part, i.e. showing how to open a span around that and an instance of the associated trace?

Cito: Ok. Maybe just mention that exploring other backends is not part of this epic but may be done later if we are not satisfied with Jaeger.

For the second part, yes, essentially just showing how to open some spans around such functions. Maybe also checking if opentelemetry-instrumentation-asyncio could be useful in this regard.

Cito (approved):

TheByronHimes (approved):
I only have questions about the role of the existing correlation ID in all of this. Seems solid to me though.

TheByronHimes (approved):

69-common-kingslayer/technical_specification.md
```diff
@@ -0,0 +1,126 @@
...
+opentelemetry-distro
+opentelemetry-instrumentation
+opentelemetry-instrumentation-logging
+opentelemetry-instrumentation-aiokafka
+opentelemetry-instrumentation-botocore
+opentelemetry-instrumentation-fastapi
+opentelemetry-instrumentation-httpx
+opentelemetry-instrumentation-pymongo
+opentelemetry-instrumentation-requests
+````
```
mephenor, line 40:
```suggestion
```
```

TheByronHimes (approved):

TheByronHimes (approved):

PR #141, opened 2025-03-10
Adapt epic docs to support multiple roles for auth (GSI-1482)

Cito:
There is only one mention of user roles in our epic docs that needs adaptation.


lkuchenb (approved):
This sounds like a very compact change 😅

PR #145, opened 2025-04-22
Access request handling improvements (Ballan Wrasse) (GSI-1560)

Cito:


73-ballan-wrasse/technical_specification.md
```diff
@@ -0,0 +1,70 @@
...
+data stewards control over the validity period and enriching the data access request with existing information (e.g., the corresponding DAC) and additional data entered
+by the data stewards (e.g., reason for rejecting a request, internal notes, ticket ID).
+
+### Included/Required
+
+- Data stewards should be able to modify the requested validity period
+  when the request is approved.
+- Data stewards should be able to add a note to the requester
+  (e.g., reason for denial of an access request or modified validity period).
+- Data stewards should be able to add a ticket ID and internal notes.
```
ac-jorellanaf, line 22: Could we make this 'ticket ID' an autoincremental number? I don't see a point in having custom IDs.

73-ballan-wrasse/technical_specification.md
```diff
@@ -0,0 +1,70 @@
...
+  access token should be shown to the user.
+
+### Not included
+
+- Collecting additional context data for the request (e.g., affiliation).
+
+## Affected services
+
+- Access Request Service:
+  - Allow changing the validity period in the PATCH request.
```
ac-jorellanaf, line 36: Should we also provide users the possibility of changing their access request access period too?

73-ballan-wrasse/technical_specification.md
```diff
@@ -0,0 +1,70 @@
...
+**Epic Type:** Implementation Epic
+
+Epic planning and implementation follow the
+[Epic Planning and Marathon SOP](https://docs.ghga-dev.de/main/sops/sop001_epic_planning.html).
+
+## Scope
+
+### Outline
+
+The goal of this epic is to improve the management of access requests by giving
```
ac-jorellanaf, line 12: Maybe we should also link the forum discussion(s) relevant to the epic?

ac-jorellanaf (commented):
All looks good. Only a couple comments/questions.

73-ballan-wrasse/technical_specification.md
Cito: Ticket ID refers to the ID that is created by the help desk where the access request is further processed. From this ID we can then derive the link to the ticket in the help desk.

Cito: I have made explained this a bit more now.

73-ballan-wrasse/technical_specification.md
Cito: They can already do that (within certain limits) when they create the request. But from then on, we should not allow further changes since that would mean that the data stewards and DAC would need to be informed about the change as well, which would complicate things a lot. Instead, any further communication should happen via the help desk.

73-ballan-wrasse/technical_specification.md
Cito: I think it's simpler to add any such links and other relevant info to the corresponding tasks in YouTrack.

73-ballan-wrasse/technical_specification.md
ac-jorellanaf: I understand, but with custom IDs, we need to check existing, process inputs, etc. A simple auto-generated auto-incremental ID (e.g. YouTrack tasks) might be better, in my opinion.

Cito: The ticket ID is generated in the help desk, so we have no control over it. We do not need to check whether it's existing etc. It just helps the data steward to find the associated help desk ticket.

ac-jorellanaf (approved):

PR #147, opened 2025-06-10
Add eurasian blackbird epic (GSI-1684)

TheByronHimes:


75-eurasian-blackbird/technical_specification.md
```diff
@@ -0,0 +1,166 @@
...
+    - Rename that `event_id` variable to `compaction_key` or similar (no side effect).
+  - DAO Pub (outbox):
+    - Here, `event_id` should only be stored, not reused during republish. The stored
+      value should reflect the ID of the event created the last time the DTO was 
+      published to Kafka, to be used for traceability only.
+      - Every time outbox events are republished, they'll have a *new* `event_id`.
+      - The outbox events are intended to be stored by consumers, so idempotence
+        is not performed via `event_id`. This is why we won't update the daosub
+        protocol or try to reuse the previous `event_id` when republishing outbox
+        events.
```
mephenor, line 77: Might need an alternative explanation for this, especially for the `stored, not reused` part. If this is replayed, wouldn't we expect a payload with an `event_id` that has been previously seen?
This would still generate the `event_id` somewhere along the way?

75-eurasian-blackbird/technical_specification.md
```diff
@@ -0,0 +1,166 @@
...
+one of those upstream services, the `NOS` will issue *new* Notification events, and
+the `NS` will send new emails. We could store the event as a whole, or store its hash
+(like in the `NS`), or store a deduplication key based on certain values, but they
+have problems:
+- *Whole event*: waste of space
+- *Hash of event*: opaque, unable to modify if we need a migration
+- *Semantic Dedupe Key*: heterogeneous format, possibility to clash
+- *All of the above*: vulnerable to upstream republish post-DB migration
+
+If all Kafka events have a standard UUID header, stored and revived like the
```
mephenor, line 45: Even though it is implicitly given, that we use UUID4, this should be explicit about that fact.

75-eurasian-blackbird/technical_specification.md
```diff
@@ -0,0 +1,166 @@
...
+
+```python
+async def consume(
+    self,
+    *,
+    payload: JsonObject,
+    type_: Ascii,
+    topic: Ascii,
+    key: Ascii,
+    event_id: UUID,
```
mephenor, line 116: Should specify the concrete UUID version, if possible

75-eurasian-blackbird/technical_specification.md
```diff
@@ -0,0 +1,166 @@
...
+
+@abstractmethod
+async def _consume_validated(
+    self,
+    *,
+    payload: JsonObject,
+    type_: Ascii,
+    topic: Ascii,
+    key: Ascii,
+    event_id: UUID,
```
mephenor, line 128: See above

75-eurasian-blackbird/technical_specification.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+## Scope
+### Outline:
+The aim of this epic is to provide Kafka event identification for deduplication by
+canonizing a UUID header for all events (`event_id`). The `NOS`, `NS`, and `DLQS` will
+receive updates targeting this change after updating `hexkit`.
+
+Changes for `hexkit` should be included in the same release as the changes from the
+`Slow Worm` epic.
```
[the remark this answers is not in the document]
TheByronHimes: It's not a dependency, it's that the territory overlaps. If we perform all of slow worm first, it will involve two major releases instead of one, and a second wave of rollout PRs too. I really suggest that we try to do them in parallel.

75-eurasian-blackbird/technical_specification.md
TheByronHimes: I'll look into it. I know pydantic has one, but I don't think there's an out-of-the-box uuid4 type (is there?). If pydantic's doesn't work, we can see about adding one to `custom_types`

75-eurasian-blackbird/technical_specification.md
TheByronHimes: Fixed

75-eurasian-blackbird/technical_specification.md
```diff
@@ -0,0 +1,166 @@
...
+deduplication store, the event will go to the DLQ and we can just give it another ID.
+
+### Included/Required:
+Hexkit:
+- Protocols:
+  - Event Pub: no change
+  - Event Sub: expose `event_id` header to translators (similar to "type_")
+- Providers:
+  - Event Sub: 
+    - Update extract `event_id` header and pass to translator.
```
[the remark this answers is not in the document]
TheByronHimes: Fixed

75-eurasian-blackbird/technical_specification.md
TheByronHimes: First, clarification: outbox events get the same event_id header, too. No difference there. They store it in the metadata field along with the correlation ID, `published` bool, etc. when saved to the DB. But we don't retrieve the value and inject it when republishing an outbox event, like we do with the persistent event publisher.

This was a difficult decision. My thoughts:
1. Passing the ID between the relevant spots in the MongoKafkaDaoPublisher code is possible but would, IMO, not be clean (modifying [this function signature](https://github.com/ghga-de/hexkit/blob/6e778b95216b01819e374132df2eea020ce98361/src/hexkit/providers/mongokafka/provider/daopub.py#L109) so we can pass the event ID [here](https://github.com/ghga-de/hexkit/blob/6e778b95216b01819e374132df2eea020ce98361/src/hexkit/providers/mongokafka/provider/daopub.py#L432)). It would also mean updating the daosub protocol to accept the event_id. Yes, the event sub protocol gets the event_id parameter, but the semantics of the two protocols are sufficiently distinct. **This point and #2 are by biggest hangup. If we collectively decide both points are no big deal, the implementation would take less time than it took to write this response.**
2. Republishing outbox events does not directly republish past events. It retrieves the stored domain objects, runs them through a function that converts them to an event, and *then* publishes them to the configured topic/type. There's more potential for something to change compared to persistent events (but this potential is still low in an absolute sense).
3. Persistent events are meant to capture a kafka event and be able to restore it should something happen to kafka. Outbox events are just domain data stored in the DB which is later converted to an event; the event itself is not stored.
4. Ensuring idempotence for outbox events is hard to mess up, to the point that restoring an event_id matching the last published version of the object is redundant.
5. You could draw a distinction between the outbox and persistent event mechanisms by saying the former republishes data while what the latter does is more akin to restoring data. They use separate protocols, after all.
6. We can add it in the future if we change our mind, too

mephenor: Thanks for the clarification, I always get the details of both DB coupled publisher types confused, mixing up compaction for the persistent publisher and the effective state of a DAO event which is/can be the result of multiple events.

mephenor (approved):

Cito (dismissed):
Sounds good.

Cito (approved):

PR #152, opened 2025-07-07
Adding screenshots / wireframes for user and access management

SilverLinings89:


78-royal-angelfish/Access_Grant.png
ac-jorellanaf, line 1: Looks good. I will just make a minor note that this is meant to be access grant management, so it should not say 'Your contract addresses', but rather 'The user's'. Same as the 'Please wait for the approval'. Doesn't seem to fit for a management window.
It's not really necessary to fix in my opinion, but I thought it would be important to point out.

ac-jorellanaf (approved):

PR #153, opened 2025-07-08
Extend scope of Royal Angelfish to adapt old UIs

Cito:


ac-jorellanaf (dismissed):
Just something I thought of that should be added

SilverLinings89 (approved):

PR #159, opened 2025-11-19
Sarcastic Fringehead: Upload Path Pt2 (GSI-1960)

TheByronHimes:
This epic details the backend work for the remaining portion of the file upload path.

The `FileUploadReport` has been augmented from the original concept. Instead of being generated by the DHFS and maintained by the FIS, it is generated with partial data by the FIS and the remaining data is supplied by DHFS after interrogation.


84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,238 @@
...
+### RESTful/Synchronous:
+
+FIS:
+- POST /federated/ingest_secret
+  - Request body must contain an encrypted file secret
+  - Returns a 201 status code and the ID of the stored secret
+- GET /uploads/{storage_alias}/inbox
+  - Returns a 200 status code and list of `FileUploads`
+- GET /uploads/{storage_alias}/archived
+  - Returns a 200 status code and a list of file_ids for deletion
```
mephenor: Confused about the deletion part here.
From the route I would expect this to provide a list of files that have been successfully moved to permanent storage.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,238 @@
...
+### Not included:
+Data Portal updates or any upcoming metadata-related services. This is purely for file upload.
+
+## User Journeys (optional)
+
+All user journeys are already detailed in Lynx Boreal. The operations added in this epic will occur automatically without further action required on the part of either the user or GHGA personnel. 
+
+## API Definitions:
+
+### RESTful/Synchronous:
```
mephenor: Need a refresher here, but is the information returned by these endpoints scoped based on claims in an authentication token or how is the visible/returnable information determined?

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,238 @@
...
+    - Updates the SHA-256 checksum over the unencrypted content
+    - Re-encrypts the chunk using the newly generated file secret
+    - Updates the MD5 checksum over the encrypted file content
+    - Uploads the re-encrypted chunk to the `interrogation` bucket
+  - Compares the unencrypted file's checksum against the one reported by the submitter during upload, and the encrypted checksum against the one calculated by S3
+  - Sends a `FileUploadReport` to the FIS's HTTP API
+
+The final duty of the DHFS is to clean up files from the `interrogation` bucket after archival. It does this by polling the FIS for a list of files that should be deleted. The FIS sources this list from the `FileInternallyRegistered` events published by the IFRS. When the DHFS deletes a file, it contacts the FIS HTTP API with the deleted file ID. This mechanism removes the need for the DHFS to keep track of the files in its `interrogation` bucket, provides idempotent behavior, and lets the DHFS carry out its own deletions (as opposed to giving the IFRS this duty). If the DHFS fails to find a file during deletion, it logs the absence but otherwise treats it as a successful deletion.
+
```
mephenor: I think this would be easier to read and clearer to understand if the FIS and DHFS descriptions swap places and try to be as centered as can be on their own descriptions.

mephenor (changes requested):

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,238 @@
...
+The new role of the FIS is to inform the DHFS when new files arrive in the DHFS's `inbox` bucket, as well as to inform the DHFS when the IFRS has finished copying successfully interrogated files to permanent storage so that they may be deleted. To perform these actions, the FIS monitors `FileUpload` outbox events published by the UCS, as well as `FileInternallyRegistered` events published by the IFRS.
+
+When a `FileUpload` event arrives, the FIS first checks to see if it has a copy of the event already stored in its database. If it does, it then looks at the timestamp; if the local copy is newer, the FIS makes a log and ignores the event. If the inbound event is newer, however, the FIS checks the state field. If the state is a valid transition, e.g. INIT to INBOX or INBOX to ARCHIVED, then the FIS updates its local copy. If the transition is not valid, then FIS raises an error and sends the event to the DLQ. If there are no problems with the event (and if it's not a deletion), the FIS serves this info to the DHFS upon request (which the DHFS will do in a polling manner).
```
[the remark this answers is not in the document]
TheByronHimes: I added a Leon-style excalidraw diagram for the main services involved and organized the content for each service better in the implementation details

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,238 @@
...
+
+The FIS operates an HTTP API with the following new endpoints:
+- a `GET` endpoint that returns a list of completed `FileUploads` for a given storage alias
+- a `GET` endpoint that returns a list of archived file IDs that should be deleted from a given storage alias
+- a `POST` endpoint that accepts `FileUploadReport` objects from Data Hubs, each of which details a file that has been [un-]successfully re-encrypted and uploaded to the `interrogation` bucket.
+- a `DELETE` endpoint that informs FIS that a given file has been removed from a given storage alias.
```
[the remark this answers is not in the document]
TheByronHimes: This is outdated now - interrogation bucket cleanup occurs via a DHFS job

84-sarcastic-fringehead/technical_specification.md
mephenor: I was thinking about whether the endpoints return all information for a storage alias unconditionally or if there's information in the token to  scope down the returned information to the corresponding research upload box.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,238 @@
...
+
+#### Work to be performed for the GHGA Connector
+- [ ] Fetch and use Data Hub public keys
+
+### FIS:
+The FIS straddles the border between the file services group and everything else, similar to the role played by the UOS. In the past, the FIS acted as a way to ingest file upload metadata and tell other services when a manually validated ("interrogated") file was ready for permanent storage. This had to be done as a temporary solution until the remote file upload and automatic file interrogation was implemented, which is the work proposed in this epic.
+
+The new role of the FIS is to inform the DHFS when new files arrive in the DHFS's `inbox` bucket, as well as to inform the DHFS when the IFRS has finished copying successfully interrogated files to permanent storage so that they may be deleted. To perform these actions, the FIS monitors `FileUpload` outbox events published by the UCS, as well as `FileInternallyRegistered` events published by the IFRS.
```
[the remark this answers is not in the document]
TheByronHimes: All of that should be clear in the specification as of now. If not, I can add further clarification somewhere.

mephenor (changes requested):
Much easier to read and well thought out version of the spec.
There's one minor and one major issue remaining:
- Box ID is mentioned in the models, but not where/when it's populated. Should be added for consistency with the other information
- The whole point where the actual data finally resides needs some more current input from the other reviewers. I think I'm operating on outdated information here, but in my mental model the data does not have to leave the data hub and only the secrets are stored at central in every case.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+
+- `POST /secrets`
+  - Authorization requires a token signed with Data Hub-specific private key
+  - Request body must contain an encrypted file secret
+  - Returns `201 CREATED`
+- `GET /uploads/{storage_alias}/inbox`
+  - Authorization requires a token signed with Data Hub-specific private key
+  - Returns `200 OK` and a list of `FileUploads` for files awaiting interrogation
+  - Files are grouped by storage alias
+- `GET /reports/{file_id}`
```
Cito: In a RESTful API, I would expect this to be called "inspection-results" instead of "reports". Consistent naming makes it a bit easier to understand this already pretty complex process.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+    DHFS->>FIS: POST InspectionResults
+    FIS->>FileUploadReports: Publish: FileUploadReport
+    FileUploadReports->>UCS: Consume: FileUploadReport
+    FileUploadReports->>IFRS: Consume: FileUploadReport
+    UCS->>inbox: Delete File
+    UCS->>FileUploads: UPSERT: FileUpload(state=ARCHIVED)
+    Note right of UCS: UCS considers items as<br> archived once they're <br> uploaded to the <br> interrogation bucket.
+    UOS->>ResearchDataUploadBoxes: UPSERT: ResearchDataUploadBox(state=CLOSED)
+    ResearchDataUploadBoxes->>IFRS: UPSERT: ResearchDataUploadBox(state=CLOSED)
+    IFRS->>interrogation: Copy File from interrogation to archive bucket
```
Cito: Maybe note that this should happen for all gathered files that are inside the box.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+- LoggingConfig
+- OpenTelemetryConfig
+- S3ObjectStoragesConfig
+- auth_token_signing_key
+  - This is the Data Hub's private key, which it uses to sign auth tokens sent to FIS
+- wkvs_api_url
+  - Used to contact WKVS to get the GHGA public key
+
+### IFRS:
+The role of the IFRS is to shepherd files into permanent storage, or "archival", by copying them from the `interrogation` bucket located at a Data Hub into the `archive` or `permanent` bucket located at GHGA Central. This only occurs once the Data Hub in question has completed the interrogation process, which is detailed in the [DHFS section](#dhfs) above. Unlike the FIS and DHFS, the IFRS operates only as an event consumer.
```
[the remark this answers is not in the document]
mephenor: So there's two things here, one on the implementation level and one on the policy level.
As for the technical part, currently there's one limitation around the storage alias, where input storage alias = output storage alias, so that would need to be changed to make it possible to move data between hubs/locations.
This probably needs updated schemas to have a notion of incoming/outgoing storage alias and might also need to change the ObjectStorages config, though not sure on the second part.
We could also decouple storage alias and location bucket ID at the same time to make this fully future proof.

This alone could already warrant a separate, self-contained epic.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+- [ ] Add outbox subscriber for `FileUpload` events
+- [ ] Add persistent publisher for `FileUploadReport` events
+- [ ] Add HTTP endpoints as outlined above
+- [ ] Write migration for existing persisted events if possible
+- [ ] Write tests
+
+
+### DHFS:
+The DHFS is a new service that is operated by the Data Hubs for the purpose of performing file validation and re-encryption, and to keep file ingest in general as a federated operation. The DHFS operates two instances: a client instance, which performs the interrogation work and is always running; and a cleanup instance, which runs at some interval and deletes files from the `interrogation` bucket once they've been copied to permanent storage.
+
```
Cito: You should mention here that the Data Hubs do not share the event streams with central, so the DFHS cannot subscribe to FileUploads directly and needs to poll a FileUpload list from central via REST  instead.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+        participant interrogation
+    end
+    
+    DHFS->>interrogation: List object IDs (polling)
+    interrogation-->>DHFS:
+    rect rgb(30, 30, 30, .8)
+        loop For each file in bucket
+            DHFS->>DINS: GET file info
+            DINS-->>DHFS: 200 | 404
+            DHFS->>interrogation: Delete file
```
Cito:
```suggestion
            DHFS->>interrogation: Delete file (if 200)
```

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+- [ ] Write migrations
+- [ ] Move migrations to own CLI command so they can be run as an init container
+  - [ ] Work with DevOps to get this configured in k8s
+
+## Diagrams:
+
+### Service Diagrams:
+#### Service Map
+![Service Map](./images/service_map.png)
+
```
Cito: Nice map. The DHFS cleanup job also GETs the inspection results from FIS - that arrow is missing.

Cito (dismissed):
Thanks. Made some comments below, but essentially looks good to me.

AC007 should be adapted afterwards to reflect the current state.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+        participant DINS
+        participant IFRS
+    end
+    box rgb(200, 75, 35, 0.5) S3 Buckets
+        participant inbox
+        participant interrogation
+        participant archive
+    end
+    
+    Connector->>UCS: Complete Upload
```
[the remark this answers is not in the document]
TheByronHimes: Please let me know if the revised version of the diagram makes more sense or if you feel it could be further clarified.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,478 @@
...
+## User Journeys (optional)
+
+All user journeys are already detailed in Lynx Boreal. The operations added in this epic will occur automatically without further action required on the part of either the user or GHGA personnel. 
+
+## API Definitions:
+
+### RESTful/Synchronous:
+
+FIS:
+> See the [FIS HTTP API](#fis-http-api) section for an in-depth explanation
```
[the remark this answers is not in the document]
TheByronHimes: I've consolidated the API definitions (or rather, the opposite) so they live with their respective service sections. The Restful/Synchronous heading now contains links to those three API definitions, and those sections in turn link back to the top for easy nav.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,742 @@
...
+Upon startup, the HTTP API instance of the FIS will retrieve the list of Data Hub public keys from the WKVS.
+
+In addition to implementing the endpoints defined here, the existing functionality and config that directly interacts with Vault should be removed so that EKSS is the sole middleman for Vault activity.
+
+> See the [diagram](#example-auth-token-structure-for-dhfs-calls-to-fis-api) for an illustration of the proposed auth token structure for inbound requests to the FIS API
+
+The FIS operates an HTTP API with these endpoints:
+1. `POST /secrets`: Accept a new file secret for deposition in the EKSS
+   - Authorization requires a token created with both the FIS public key and the Data Hub-specific private key
+     - Token is should contain a sub-token encrypted with the Data Hub's private key and `storage_alias` and be encrypted with FIS public key.
```
mephenor:
```suggestion
     - Token should contain a sub-token encrypted with the Data Hub's private key and `storage_alias` and be encrypted with FIS public key. 
```

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,742 @@
...
+
+In addition to implementing the endpoints defined here, the existing functionality and config that directly interacts with Vault should be removed so that EKSS is the sole middleman for Vault activity.
+
+> See the [diagram](#example-auth-token-structure-for-dhfs-calls-to-fis-api) for an illustration of the proposed auth token structure for inbound requests to the FIS API
+
+The FIS operates an HTTP API with these endpoints:
+1. `POST /secrets`: Accept a new file secret for deposition in the EKSS
+   - Authorization requires a token created with both the FIS public key and the Data Hub-specific private key
+     - Token is should contain a sub-token encrypted with the Data Hub's private key and `storage_alias` and be encrypted with FIS public key. 
+     - FIS decrypts the outer token to learn which Data Hub public key to use to decrypt the inner token. The inner token certifies that the request was indeed sent from the given Data Hub.
```
mephenor:
```suggestion
     - FIS decrypts the outer token with its private key to learn which Data Hub public key to use to decrypt the inner token. The inner token certifies that the request was indeed sent from the given Data Hub.
```

For added clarity

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,742 @@
...
+- **One alternative** is to manually exfiltrate the data to UCS.
+- **A second alternative** is to simply drop the data since the files have already been archived and all relevant information is actually stored by IFRS.
+
+There is one **problem**: IFRS has different object IDs than FIS. If we keep the event data which FIS currently has, we will need to update the information so that the file IDs in FIS are set to the file IDs (object IDs) known to IFRS, using the accession to match. That action would not be reversible, of course.
```
mephenor: The failure mode here would be IFRS loosing information in its internal DB about files that are already archived.

Theoretically we should preserve the information to keep guarantees about what we are able to restore by replaying events, even though personally I'd prefer a clean cutoff as there's a separation between current and new FIS which, in my mind, make them two distinct services and the new one shouldn't be burdened by one off logic permanently in it's code.
Would it make sense to possibly keep this legacy migration separate as a one off script?

On another note, why does this have to be reversible?

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,742 @@
...
+> [Return to API list](#restfulsynchronous)
+
+The UOS would get the following new endpoint:  
+`PATCH /boxes/{box_id}/accessions`: Accept a file-to-accession mapping in order to relay it to the UCS
+- Authorization uses auth context protocol and requires a Data Steward role
+- Request body must contain a payload conforming to the `FileAccessionMap` [schema](#fileaccessionmap)
+- Returns `204 NO CONTENT`
+- Description:
+  - UOS looks for the `ResearchDataUploadBox` in its database with an ID that matches the value in the path parameter, and returns a `404 NOT FOUND` if it doesn't find it.
+  - UOS ensures the `ResearchDataUploadBox` is not already CLOSED, and raises an error if it is. This might return a `409 CONFLICT` status code.
```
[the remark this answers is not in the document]
TheByronHimes: Good catch, looks like I missed a spot when changing that to ARCHIVED.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,742 @@
...
+1. `POST /secrets`: Accept a new file secret for deposition in the EKSS
+   - Authorization requires a token created with both the FIS public key and the Data Hub-specific private key
+     - Token is should contain a sub-token encrypted with the Data Hub's private key and `storage_alias` and be encrypted with FIS public key. 
+     - FIS decrypts the outer token to learn which Data Hub public key to use to decrypt the inner token. The inner token certifies that the request was indeed sent from the given Data Hub.
+     - Inner token, signed by the Data Hub private key, contains the file ID
+   - Request body must contain the associated file ID and a file secret encrypted with the GHGA central public key
+   - Returns `201 CREATED`
+   - Description:
+     - FIS finds the existing `FileUnderInterrogation` in its database, raising an error if it doesn't find it.
+     - FIS forwards the file secret to EKSS (still encrypted) in exchange for a secret ID.
```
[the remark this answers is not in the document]
TheByronHimes: The DHFS should encrypt the actual secret with its own private key, then encrypt that information and the data hub ID with the FIS public key. The FIS should decrypt the outer layer, fetch the public key of the Data Hub, then decrypt it and re-encrypt it with the GHGA public key before sending it to the EKSS. That way the EKSS doesn't need to know anything about Data Hubs.

84-sarcastic-fringehead/technical_specification.md
TheByronHimes: Good point. Take a look at the new wording in [818fac5](https://github.com/ghga-de/epic-docs/pull/159/commits/818fac5db25fb2b7a65ddf294bb002124f9dabff) and let me know if that sounds better.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,774 @@
...
+# Additionally, the following fields exist but are unset until later in the process
+secret_id: str | None
+decrypted_sha256: str | None  # SHA-256 checksum of the entire unencrypted file content
+decrypted_size: int     # The size of the unencrypted file
+part_size: int  # The number of bytes in each file part (last part is likely smaller)
+encrypted_parts_md5: list[str] | None     # Is None until DHFS finishes with file
+encrypted_parts_sha256: list[str] | None  # Is None until DHFS finishes with file
+accession: str | None   # The accession number for the file
+```
+
```
Cito: Add a note that the corresponding mongodb collection should have a compound unique indexes for `(box_id, alias)`.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,774 @@
...
+```python
+id: UUID4       # Unique identifier for the file upload
+box_id: UUID4   # The ID of the FileUploadBox this FileUpload belongs to
+alias: str      # The submitted alias from the metadata (unique within the box)
+state: FileUploadState = "init"  # The state of the FileUpload
+state_updated: UTCDatetime  # Timestamp of when state was updated
+storage_alias: str  # The storage alias for the inbox bucket
+# Additionally, the following fields exist but are unset until later in the process
+secret_id: str | None
+decrypted_sha256: str | None  # SHA-256 checksum of the entire unencrypted file content
```
Cito: For completeness, should we also have a `decrypted_md5`?

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,774 @@
...
+The UCS operates the following new endpoints:
+- `PATCH /boxes/{box_id}/accessions`: Accepts a payload which maps `FileUpload` IDs to accession numbers for a given `FileUploadBox`
+  - Authorized by a token signed by the UOS with the work type `"map"` and including the `box_id`
+  - Returns `204 NO CONTENT`
+  - Description:
+    - The UCS finds the `FileUploadBox` in its database and raises an error if it can't
+    - The UCS ensures the `FileUploadBox` is locked but not archived, raising an error otherwise.
+    - Before making database writes, the UCS retrieves all specified `FileUpload` objects and:
+      - Ensures the file state is not `archived`.
+      - Ensures the file accession field is not assigned, or if it is assigned, that it matches the accession provided in the request. Otherwise, UCS raises an error and returns a `409 CONFLICT` response.
```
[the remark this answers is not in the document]
Cito: Shouldn't this also check that the accessions are unique (see above)?

Cito (approved):
Made some suggestions, but I have no major objections.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,774 @@
...
+    note left of inbox
+      Source of truth:
+      - State set by UCS
+      - Transitions driven by InterrogationSuccess/InterrogationFailure published by FIS
+    end note
+```
+
+#### Example Auth Token Structure for DHFS calls to FIS API
+> See the [FIS HTTP API](#fis-http-api) section for context.
+
```
Cito: As discussed, could also be a JWT signed by the respective hub.

Cito (dismissed):

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,742 @@
...
+```
+
+#### InterrogationFailure
+> Persistent event published by FIS on behalf of DHFS
+```python
+file_id: UUID4
+storage_alias: str  # the interrogation bucket storage alias
+interrogated_at: UTCDatetime
+reason: str
+```
```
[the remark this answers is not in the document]
TheByronHimes: @mephenor I chatted with @Cito briefly and he seems okay with allowing these concepts in the UCS because it's taking on a larger role in this epic than initially defined in Lynx Boreal.

84-sarcastic-fringehead/technical_specification.md
TheByronHimes: I have updated the document for this change. Please let me know if what I wrote matches your suggestion!

Cito (changes requested):
See comment regarding passing file IDs as query parameters.

84-sarcastic-fringehead/technical_specification.md
```diff
@@ -0,0 +1,763 @@
...
+  - Sends an `InterrogationReport` to the FIS's HTTP API
+    - In the successful case, the `InterrogationReport` includes the new file encryption secret encrypted with the GHGA public key.
+
+#### DHFS Cleanup Job (secondary instance)
+The secondary duty of the DHFS is to clean up files from the `interrogation` bucket. Files must be removed once they have been fully copied to the permanent bucket, as well as on occasions that files are deleted from their parent box. Neither the FIS, UCS, nor IFRS can perform this action because they don't have write access to the `interrogation` bucket. Each time this DHFS instance runs, it retrieves a list of all objects (files) currently in the `interrogation` bucket. Then the DHFS makes a single GET request to the FIS API's `GET /storage/{storage_alias}/uploads/can_remove` endpoint and supplies the file IDs as query parameters (`file_id=<file_id>&file_id=<file_id>,...`). For authentication, the DHFS signs a JWT with its private key. In response, the DHFS expects to get a list containing the IDs of files which may be deleted from the interrogation bucket. The DHFS will then *delete* each listed file from the `interrogation` bucket.
```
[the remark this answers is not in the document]
TheByronHimes: Updated

Cito (dismissed):

mephenor (dismissed):

sbilge (approved):

PR #163, opened 2026-02-18
Initial GHGA Registry Service (Apollo)

Cito:
The Apollo epic aims at creating a first working implementation of the GHGA Registry Service.


ckaipf (dismissed):

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+
+The second crucial change introduced with the SRS is the separation of Experimental Metadata from Persistent Administrative Metadata, Dynamic Administrative Metadata, and Datasets, which were previously bundled into an all-in-one [GHGA metadata schema](https://github.com/ghga-de/ghga-metadata-schema), and the move from the former dataset-centric metadata concept to a study-centric one.
+
+The Experimental Metadata should be managed in a generic way, without semantic interpretation of its content besides generating an accession number for each entity.
+
+As a reminder, Experimental Metadata (EM) is data describing how the research data (that we store in “files”) was generated. Persistent Administrative Metadata (PAM) is non-experimental study metadata (e.g. authors, description), and Dynamic Administrative Metadata (DAM) is business metadata (e.g. authorized users, DAC contacts and policies). As the names indicate, PAM underlies the same archival immutability as the EM, while DAM is auditable.
```
sbilge:
```suggestion
As a reminder, Experimental Metadata (EM) is data describing how the research data was generated. Persistent Administrative Metadata (PAM) is non-experimental study metadata (e.g. authors, description), and Dynamic Administrative Metadata (DAM) is business metadata (e.g. authorized users, DAC contacts and policies). As the names indicate, PAM underlies the same archival immutability as the EM, while DAM is auditable.
```

sbilge: The “files” we store don’t exactly correspond to the research data represented in the metadata model. In the current model, we distinguish between submitted file types (e.g., research data files, processed data, analysis support files, etc.). I removed the term to eliminate jargon ambiguity between the model perspective and the backend perspective.

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+
+An Experimental Metadata Ingress Model (EMIM) shall be used for validation. EMIMs shall be stored as schemapacks, and the EM shall be stored as datapacks. Both structures are provided by the [schemapack](https://github.com/ghga-de/schemapack) library.
```
sbilge:
```suggestion
An Experimental Metadata Ingress Model (EMIM) shall be used for validation. EMIMs shall be stored as schemapacks, and the EMs shall be stored as datapacks. Both structures are provided by the [schemapack](https://github.com/ghga-de/schemapack) library.
```

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+
+All resources archived in GHGA that are subject to archival immutability can be retrieved by an accession number, a persistent identifier (PID) that always resolves to the exact same state of a resource.
```
sbilge: Should we detail what we mean by the resources?

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+
+- `id: str` - the code of the DAC (short, uppercase) (primary key)
+- `name: str` - human readable name of the DAC
+- `email: EmailStr` - the contact email address of the DAC
+- `institute: str` - the institute the DAC belongs to
+- `created: Date` - when the DAC entry was created
+- `changed: Date` - when the DAC entry was last changed
+- `active: bool` - whether the DAC is still active
+
+Note: The `name` should correspond to the `alias` in the old model. The `id` should be derived from the name (shortened, converted to upper case). We do not assign a citable accession number to DACs anymore.
```
sbilge: Do you mean the `name` or the `id` should correspond to the `alias`?

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+- `changed: Date` - when the DAP entry was last changed
+- `active: bool` - whether the DAP is still active
+
+The `id` should correspond to the `alias` in the old model, `text` and `url` correspond to `policy_text` and `policy_url`. We do not assign a citable accession number to DAPs anymore.
+
+New DataAccessPolicy entity instances should only be created after verification that the corresponding DataAccessCommittee entity instance exists.
+
+#### Dataset
+
+The Dataset entity describes a set of files and represents the smallest unit for which a data access request can be formulated. All attributes except `dap_id` are immutable, and Dataset entity instances cannot be deleted unless the study still has the status `PENDING`. However, new Dataset entity instances can be created after study creation and assigned to a study without violating its immutability. An immutable accession number is assigned upon creation. Every Dataset belongs to exactly one Study. The DataAccessPolicy assignment is mutable.
```
sbilge: Isn't this the same as the dataset-centric approach we currently have? In the study-centric approach, is the link between a study and a file still a dataset?

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+- `resource: TypedResource` - the kind of resource this type belongs to
+- `name: str` - the human readable name of the resource type
+- `description: str | None` - optional definition text or help text
+- `created: Date` - when the resource type was created
+- `changed: Date` - when the resource type was last changed
+- `active: bool` - whether the resource type is still active
+
+The corresponding collection should be created with a composite index on `code` and `resource`.
+
+The corresponding collection can be populated with the study types as defined in the existing GHGA metadata schema. The existing dataset types need to be extracted from the current submission store, as they are not defined in the existing schema.
```
sbilge:
```suggestion
The corresponding collection can be populated with the study types as defined in the existing GHGA metadata schema (StudyTypeEnum). The existing dataset types need to be extracted from the current submission store, as they are not defined in the existing schema.
```

sbilge: “Dataset type” is currently a free-text field. Values from past submissions will not reflect the full range of dataset types that could appear in future submissions.

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+The service will then create an AnnotatedEMPack and publish it as an event if the exact same AnnotatedEMPack has not been published before, as described further below.
+
+The service also supports mapping uploaded files to their experimental metadata entries. See the `POST /file-ids` endpoint below.
+
+### Validation
+
+On ingress, submissions must be validated and rejected if there are any validation errors. Validation happens on several levels:
+
+- ingested DAM and PAM (e.g., study or dataset) are validated as DTOs using Pydantic
+- the submission must be complete (particularly, it must have a Publication and ExperimentalMetadata)
```
sbilge: Not all submissions have a publication. Some users deposit their data before their publication is accepted. Will the service accept an empty Publication object?

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+If study submissions contain non-public metadata, this metadata must be provided as files contained in a dataset that needs to be requested like other datasets.
+
+## User Journeys
+
+### New Study
+
+Typical user journey for a data steward creating a new study:
+
+- data steward logs into the data portal
+- submits a new study type via `POST /resource-types` if needed
```
sbilge, line 444: Does it mean that the enum values of a controlled vocabulary will be extended by one data steward responsible for the submission?

87-apollo/technical_specification.md
Cito: I agree, it could be misleading since metadata contains not only info about the generation of the data, and the files not only contain research data in the narrow sense. But I would still like to mention files in this sentence to point out the dichotomy between the actual data (in files) and metadata.

Would this be better?

“Experimental Metadata (EM) is structured information describing the origin, context, methods, and relationships of the experimental data files (both primary and derived data) without containing the data itself.”

87-apollo/technical_specification.md
Cito: How would you define it? Would this work? “A resource is a discrete, identifiable unit of data or metadata archived by GHGA, represented either as a raw file or as a structured metadata object.”

87-apollo/technical_specification.md
Cito: The word "correspond" was a bit unclear, rephrased it in 1986ac6.

87-apollo/technical_specification.md
Cito: I now wrote "EM objects" since EM can't be pluralized.

87-apollo/technical_specification.md
Cito: A dataset is just a subset of all files mentioned in the EM (which now belongs to a study). A study can now have multiple datasets, without repeating the EM in every dataset. I rephrased this a bit in 998e111.

87-apollo/technical_specification.md
Cito: The study types had been taken over from EGA, as a fixed enum in the model, but the dataset types had been left as free text because there were some issues with their definition in the EGA model. Both approaches were not really optimal.

Data stewards can now supplement both dataset types and study types via the `POST /resource-types` endpoint if needed. I think this is a good compromise between an enum and a free text.

87-apollo/technical_specification.md
Cito: I don't think we should use a dummy object in this case, since publications also fall under archival immutability, and it would be already shown in the portal. Instead, we should then probably allow submission without a publication (just showing a warning if you try to do that), and allow adding the publication at a later date. Would that work?

87-apollo/technical_specification.md
Cito: Yes, but only for the dataset and study types which are under our control and not part of the EM.

87-apollo/technical_specification.md
lkuchenb: The official GHGA term for the files is "research data". The fact that the metadata model currently has a class "research data files" among other files doesn't help, but I would still stick to the "research data" terminology since it's the most established.

87-apollo/technical_specification.md
lkuchenb: > Isn't this the same as the dataset-centric approach we currently have? In the study-centric approach, is the link between a study and a file still a dataset?

yes

87-apollo/technical_specification.md
lkuchenb: I would suggest to skip the API controlled enums for the time being and make them enums which are configurable on the service level.

87-apollo/technical_specification.md
lkuchenb: Yes, we need to allow for studies to not have a publication.

87-apollo/technical_specification.md
lkuchenb: see above, I don't think this is needed for now

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+**Epic Type:** Implementation Epic
+
+Epic planning and implementation follow the
+[Epic Planning and Marathon SOP](https://ghga.pages.hzdr.de/internal.ghga.de/main/sops/development/epic_planning/).
+
+## Scope
+
+### Outline
+
+The Study Registry Service (SRS) is the core GHGA registry service responsible for ingesting and archiving metadata from data submitters.
```
lkuchenb: Given that also study-independent entities such as DAC and DAP are managed by this service, I tend towards renaming the entire service as GHGA Registry Service. Let's discuss.

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+- REST API as detailed below
+- Event publisher as detailed below
+
+### Not included
+
+The following features are not included in the first version of the study registry:
+
+- Migration of the existing submission store to the Study Registry Service.
+- Change requests to existing studies. In the future, it should be possible to create a modified copy of an existing study.
+- New PID schema.
```
lkuchenb: Since we need to implement the accession generator in the scope of this epic I would suggest we in fact to transition to the new PID schema, otherwise we have to implement the accession generation logic twice. Some thoughts here:

* The entire service / client stack needs to be agnostic regarding the structure of accessions and be able to process arbitrary strings
* The previously issued accessions (GHGA[A-Z][0-9]{14}) will be kept for already accepted studies and newly accepted studies as well as future revisions of those early studies will follow the new schema

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+- Support for multiple or versioned EMIMs, integration of an EMIM registry.
+- Frontend to ingest submissions and manage dynamic administrative metadata and lookups.
+- Populating and updating lookup tables from existing dictionaries and ontologies.
+- Implementation of the EM validation service.
```
lkuchenb: The service needs to validate the data upon ingress though, assuming this will be happening directly in the service?

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,810 @@
...
+
+Attributes:
+
+- `id: str` - the PID of the study (primary key)
+- `title: str` - comprehensive title for the study
+- `description: str` - detailed description (abstract) describing the goals of the study
+- `types: list[StudyType]` - the type(s) of this study
+- `affiliations: list[str]` - the affiliation(s) associated with this study
+- `status: StudyStatus` - the current status of the study (see below)
+- `users: list[UUID] | None` - user(s) who can access the study (None means publicly accessible)
```
lkuchenb: Shouldn't this be handled by CRS?

lkuchenb (commented):
Thanks for the spec, great work!

Some things I would like to discuss in the meeting tomorrow:

* Why the temporary auth concept, is there a major overhead to go directly to the CRS from the start?
* Why do we need the pre-publish state? I could imagine validation to happen immediately at POST, i.e. no invalid data to be ever accepted. Data can be immediately processed and previewed as updates progress.
* Publications must be entirely dynamic with no ties to the study lifecycle other than that the corresponding study must exist (in any state). They also do not require an accession number, citing a citation makes little sense, was more of an artifact of embedding them in the metadata

Something else I realized we should be thinking about rather sooner than later is the global uniqueness of identifiers across all classes. Currently our ingress model does not require that (thus you have to specify class + identifier). Post-accession assignment this is currently guaranteed (yet we don't make use of it at the API level). Also, the idea was to transfer the user-specified aliases from the ingress data directly to the GHGA accessions in the future, which opens the question in how far we have to transition to making this a constraint in the ingress data.

87-apollo/technical_specification.md
Cito: closing this as resolved since after another discussion, we decided to keep it

87-apollo/technical_specification.md
Cito: Changed, but let the exact open as TBD.

87-apollo/technical_specification.md
Cito: We decided to make the EMTS the authoritative source of EMIMs for now. Left it as an implementation detail whether the service fetches the model from EMTS and validates itself or lets the EMTS validate.

87-apollo/technical_specification.md
Cito: I wanted to keep it simple in the beginning, but after some discussion, we decided to already hand this off to the CRS.

Cito:
Thanks for all the great input so far here and in separate discussions. I updated the epic spec accordingly for another round of review.

87-apollo/technical_specification.md
```diff
@@ -0,0 +1,1002 @@
...
+  - `version: int` - the current RDUB version
+  - `mapping: dict[str, UUID]` - mapping from file accessions to internal file IDs
+  - `study_id: str` - the ID of the corresponding study
+- Returns: 204 or error code
+
+Ingests a filename mapping for the Research Data Upload Box with the given ID.
+
+As a safety measure, the SR must verify that all accessions in the mapping belong to the study with the specified PID and that all file IDs in the mapping belong to the box with the given ID, and respond with an error code 409 otherwise.
+
+The service should then upsert an `AltAccession` instance with type `FILE_ID` for all entries in the passed map, where `pid` is the key and `id` is the value in the map.
```
TheByronHimes: I thought the `AltAccession` events were published as outbox events. If this is true, then this will not work as currently arranged. The `AltAccession` is stored with `id` as the primary key, which means the event `key` will be set to `id` by hexkit.

TheByronHimes (changes requested):

87-apollo/technical_specification.md
Cito: Oh, that was mistakable: With key I meant "key in the map", not "event key". Do you think it is a problem if the alt accession (i.e. the file id that is a UUID) becomes the event key? If no, we can keep it like that, and I would just make this paragraph more clear.

sbilge (dismissed):

87-apollo/technical_specification.md
TheByronHimes: This should now reference the `FileAccessionMapping` which exists in `ghga-event-schemas`.

TheByronHimes (changes requested):

87-apollo/technical_specification.md
Cito: I have adapted the spec now, can you review again?

87-apollo/technical_specification.md
```diff
@@ -210,7 +212,21 @@ Attributes:
...
+
+The corresponding collection should have a unique composite index on `id` and `type`.
+
+### FileAccession
+
+The FileAccession entity stores all existing File accessions together with their corresponding internal file IDs.
+
+Attributes:
+
+- `id: UUID4` - the internal file ID (primary key)
```
TheByronHimes: If the current implementation of `FileAccessionMapping` in the WPS, DINS, and RS goes live before this is implemented, we will need migrations for all three of those services. The current naming is `file_id` and `accession` without a `created` field. What should we do for files that already exist wrt the `created` field? Just use the upload date?

Cito: Yes, the upload date would be good. But do they really need the created field? I think it would be sufficient if the RS as the owner knows the creation date.

TheByronHimes: If it's RS-only and not part of the shared/broadcasted schema, then I don't have a problem with that.

sbilge (approved):

PR #164, opened 2026-02-19
(Question Mark Chrysalis) Add information on proposed update/lock mechanism

mephenor:


Cito (dismissed):
This should work. Made some suggestions.

Cito (approved):

PR #165, opened 2026-04-02
[Question Mark Chrysalis] Added description for DB queue/poll/process loop

mephenor:


TheByronHimes (dismissed):

sbilge (commented):
The figure under “Document State Lifecycle” is not fully rendered—the descriptions within the parentheses appear to be cut off, which affects readability.

Regarding new versions of the AEMPack: I recall a discussion about using a parameter (the event key, was it?) to identify when an event is a newer version of a previously published one. It would be helpful to include that detail here.

mephenor:
> The figure under “Document State Lifecycle” is not fully rendered—the descriptions within the parentheses appear to be cut off, which affects readability.
> 
> Regarding new versions of the AEMPack: I recall a discussion about using a parameter (the event key, was it?) to identify when an event is a newer version of a previously published one. It would be helpful to include that detail here.

@sbilge 
Can't see the issue on my side.
I used <br \\> for linebreaks, IIRC this might be interpreted differently in Safari.
Can you send a screenshot?

Regarding the new versions: For this case we can treat the event key as an implementation detail of the Kafka pubsub process.
There's no additional information in there that's not also covered by the field used as primary key for the resulting document.
Currently this unconditionally flags for reprocessing, but if we want to save some work we would need to compare documents.

sbilge (approved):

PR #166, opened 2026-04-15
Upload size limits (GSI-2298)

TheByronHimes:


88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+- **Concurrent upload session cap per box:** Similar to the box limit, this would count the number of in-progress uploads and reject any requests to create new FileUploads if a configured limit would be crossed. Note that while box size limits are specific to the box, the concurrent uploads per box cap applies uniformly to all boxes. This concurrent upload cap would prevent the style of abuse where thousands of uploads are opened simultaneously, but it does not prevent the scenario where many very small uploads are created and completed in rapid succession. Note that as it's currently implemented, the GHGA Connector only uploads files in sequence anyway, meaning this limit won't be hit unless the Connector is modified.
+- **Stale session TTL with automated abort:** Track session activity via a KV store entry (key = `file_id`, value = timestamp of last activity), set on session creation and refreshed on every presigned URL issuance. Implement a periodic cleanup job that aborts S3 multipart uploads and marks `FileUpload` records as `cancelled` for sessions whose KV store entry has expired. Make this functionality accessible through a new entrypoint command.
```
mephenor: Without rate limiting presigned URL issuance it might be a good idea to delegate the update to a an async task, so the main functionality isn't blocked by another IO bound call.

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Optional:
+- **Rate limiting on presigned URL issuance:** Apply a per-file-upload token bucket to the `GET .../parts/{part_no}` endpoint to cap the rate at which a single upload session can request presigned URLs. The token bucket can be held in-memory or backed by `hexkit`'s KV store / MongoDB provider.
+- **Verify part sizes retrospectively:** Call the S3 list_parts function for the previous part (`n-1`) when a URL is requested and aborts the upload if the part size is too large. When the user tries to request the next URL or to complete the upload, they will receive an error indicating the upload has been aborted (`state="cancelled"`).
```
mephenor: This needs slight rephrasing to clarify the order of events a bit.
Does this all happen as a response to one call or is this split between two calls, aborting happening in the first one and rejection in the second one?

mephenor: Part size too large also needs some clarification.
What exactly does this mean?

S3 should give us guarantees that the all parts, except for the last one, have equal size and the last part is either equal or smaller, so if it we need to check part size, we can just check the first one when uploading the first part and everything that's somehow larger, should be rejected automatically by the S3.

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+
+The library version of `FileUploadBox` and `ResearchDataUploadBox` in `ghga-event-schemas` must be updated with the new `max_total_bytes` field. Otherwise, no new event schemas will be required and no changes to the library version of the `FileUpload` model will be needed. Session activity will be tracked in the KV store (see below), not in the outbox-published document.
+
+### Configuration:
+
+The following new config fields will be added to UCS config. All will have safe defaults:
+
+```python
+# Concurrent session cap
+max_concurrent_uploads_per_box: int = 5
```
mephenor: Should add a description to the config field that this is a global limit for all boxes.

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+
+# Stale session TTL
+upload_session_ttl_hours: int = 72             # Sessions idle beyond this are aborted
+cleanup_interval_minutes: int = 60             # How often the cleanup job runs
+
+# Presigned URL rate limiting (optional)
+max_part_urls_per_minute: int = 0      # 0 = disabled; token bucket refill rate
+part_url_burst_size: int = 10                  # Token bucket initial/max tokens
```
mephenor: Would go for a slightly different name here to make clear what that means in isolation, i.e. how many presigned URLs per timespan.
Might also be worth to expose the time span separately instead of hardcoding it to a minute to better fine tune spacing of requests.

One minute might be a long enough time span to exhaust the token bucket at the very beginning and then wait for the rest.
Would be better if we can test and fine tune with different bucket sizes/time spans to condition that behavior.

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+3. Repeats.
+
+_Alternatively, this can be run as a one-off job scheduled as a cron job._
+
+For each configured `inbox` bucket (i.e. each data hub), do the following:
+- Fetch all `state="init"` FileUploads from the DB.
+- Build a list of upload IDs for FileUploads whose last activity timestamp is older than the configured cutoff.
+- Extend the list with any uploads where an activity timestamp is missing but `FileUpload.initiated` is older than the configured cutoff (no parts complete, presumed stale).
+- Get all ongoing uploads from the bucket.
+- Extend the list with any upload IDs that don't correspond to any `FileUpload.object_id` from the first DB fetch.
```
mephenor, line 118: How do we get to that state, edge cases that exist on the S3 side, but crashed before `state=init` was set successfully?

88-pistol-shrimp/technical_specification.md
TheByronHimes: This is the scenario where someone uses a presigned URL to upload more data. For example, if the agreed file size is ~1 GiB split across 10 parts a little over 100 MiB each, but then they upload the maximum 5 GiB for each part instead. That would be rejected on completion, but they could intentionally let it hang until the upload reaches the configured expiration time and gets cleaned up. This would let us reject it right away after the first outsized part was uploaded. There isn't any way to embed size limits in the presigned URL itself for multipart uploads.

In theory the potential for abuse is low due to all of the hoops involved in someone obtaining upload access in the first place, and this is an aggressive countermeasure, which is why it's optional.

88-pistol-shrimp/technical_specification.md
TheByronHimes: Most likely case is a hard crash between MPU init and storing the FileUpload in the DB. This shouldn't happen, but if it does then the job will clean it up.

88-pistol-shrimp/technical_specification.md
TheByronHimes: I've adjusted it and think it fits better now.

88-pistol-shrimp/technical_specification.md
TheByronHimes: I've reworded the passage from your first comment.

88-pistol-shrimp/technical_specification.md
mephenor: Still not fully sold on that.

What about instead separating the time span an refill_rate per time span, so they are independently configurable and can be adapted to changing circumstances?

88-pistol-shrimp/technical_specification.md
```diff
@@ -12,11 +12,11 @@ The Upload Controller Service (UCS) currently has verification and auth mechanis
...
 
 ### Optional:
 - **Rate limiting on presigned URL issuance:** Apply a per-file-upload token bucket to the `GET .../parts/{part_no}` endpoint to cap the rate at which a single upload session can request presigned URLs. The token bucket can be held in-memory or backed by `hexkit`'s KV store / MongoDB provider.
-- **Verify part sizes retrospectively:** Call the S3 list_parts function for the previous part (`n-1`) when a URL is requested and aborts the upload if the part size is too large. When the user tries to request the next URL or to complete the upload, they will receive an error indicating the upload has been aborted (`state="cancelled"`).
+- **Verify part sizes retrospectively:** Call the S3 list_parts function for the previous part (`n-1`) when a URL is requested. If the part size exceeds the expected part size by some allowable buffer, UCS aborts the upload immediately. UCS then responds to the HTTP request with an error indicating the upload has been aborted (`state="cancelled"`). The GHGA Connector won't make subsequent requests and will display an appropriate message to the user.
```
mephenor, line 19: So this is here to allow for automatic/necessary part size adjustments, while rejecting massive part sizes that would exceed the allotted quota?

What's an appropriate buffer here, roughly 2 orders of magnitude, adjusted for the remaining quota?

88-pistol-shrimp/technical_specification.md
TheByronHimes: Per our offline discussion, we're using a refill interval param

88-pistol-shrimp/technical_specification.md
TheByronHimes: FileUpload creation requires the Connector to submit the `part_size` parameter, which should be [adjusted in advance](https://github.com/ghga-de/ghga-connector/blob/f88e457a9c7eb7649bd9911e367160d94dd36470/src/ghga_connector/core/uploading/structs.py#L87).

So I think a fairly small buffer could make sense, something like 1%?

mephenor: Ok, so this is only for the auto-adjustment, accounting for encryption details.
Let's keep it an implementation detail then

mephenor (dismissed):

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+- **Concurrent upload session cap per box:** Similar to the box limit, this would count the number of in-progress uploads and reject any requests to create new FileUploads if a configured limit would be crossed. Note that while box size limits are specific to the box, the concurrent uploads per box cap applies uniformly to all boxes. This concurrent upload cap would prevent the style of abuse where thousands of uploads are opened simultaneously, but it does not prevent the scenario where many very small uploads are created and completed in rapid succession. Note that as it's currently implemented, the GHGA Connector only uploads files in sequence anyway, meaning this limit won't be hit unless the Connector is modified.
```
Cito: Users could run multiple Connector instances in parallel though.

88-pistol-shrimp/technical_specification.md
Cito: Why can't we just use the TTL provided by S3 for presigned URLs?

Cito: LIkewise, can't we make use of [condition matching](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html#sigv4-ConditionMatching) for the content length?

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+- [ ] Add `cleanup_stale_uploads()` to `UploadControllerPort` and `UploadController`
+- [ ] Add `run_cleanup_job()` entrypoint to `main.py`
+
+---
+
+### UCS — Rate limiting on presigned URL issuance (Optional)
+
+If implemented, a per-file-upload token bucket will be applied in `UploadController.get_part_upload_url()` before the S3 presign call. Each bucket will start at `config.max_url_buildup` tokens. Once the count falls below maximum it will be incremented once every `config.part_url_refill_interval_ms` milliseconds until it reaches the maximum again. On each URL request, one token will be consumed; if the bucket is empty, `PartUrlRateLimitError` will be raised. If `config.part_url_refill_interval_ms == 0`, the feature will be disabled entirely, letting URLs be requested as fast as infra allows.
+
+A new error class will be defined: `PartUrlRateLimitError`, translated in the HTTP response as `429 Too Many Requests` with the `retry-after` set to `config.part_url_refill_interval_ms`.
```
Cito: Note that retry-after is in seconds, so needs to be converted.

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Additional Implementation Details:
+
+### UCS and RS — New upload validation
+
+In `UploadController.initiate_file_upload()` (`core/controller.py`), after the box is validated and before `S3ClientPort.init_multipart_upload()` is called, the following checks will be added:
+
+1. **Part count cap:** Will raise `PartCountLimitExceededError` if part count exceeds 10,000.
+2. **Box aggregate size cap:** Will sum `decrypted_size` across all `state="init"` FileUploads for the box (a single `find_all` query, analogous to `_update_box_stats()`). If the sum plus the new file's `decrypted_size` would exceed `box.max_total_bytes`, will raise `BoxSizeLimitExceededError`.
+3. **Concurrent upload cap:** Will count the `state="init"` FileUploads already fetched for the box aggregate check. If the count is already at `config.max_concurrent_uploads_per_box` (when > 0), will raise `TooManyConcurrentUploadsError`.
```
Cito, line 74: The counter would need a lock to avoid race conditions.

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+- [ ] Map to HTTP 429 in the FastAPI adapter
+- [ ] Add tests for: below-limit success, burst allowance, sustained rate enforcement, limit disabled when config is 0
+
+---
+
+### UCS — Verify part sizes retrospectively (Optional)
+
+> **Prerequisite:** This feature requires `hexkit`'s `ObjectStorageProtocol` to expose a `list_parts()` method.
+
+If implemented, a retrospective size check will be added in `UploadController.get_part_upload_url()` before the S3 presign call. When `part_no > 1`, `S3ClientPort.list_parts()` will be called to retrieve the metadata for part `n-1`. If the returned part size exceeds `FileUpload.part_size`, the multipart upload will be immediately aborted via `S3ClientPort.abort_multipart_upload()`, the `FileUpload` will be marked `cancelled`, and an upload-cancelled error will be raised so the caller receives a meaningful response. Subsequent requests for this upload will also fail with the cancelled-state error via the normal state check.
```
Cito: Which status code should the error have?

88-pistol-shrimp/technical_specification.md
```diff
@@ -0,0 +1,171 @@
...
+
+## API Definitions:
+
+### RESTful/Synchronous:
+
+No new endpoints will be added. The following existing endpoints will gain new validation behavior and may return new error responses:
+
+**`POST /boxes/{box_id}/uploads`** -> Initiate a new file upload session:
+- Will return `400 Bad Request` (`PartCountLimitExceededError`) if the computed part count exceeds the S3 hard limit of 10,000 parts.
+- Will return `409 Conflict` (`BoxSizeLimitExceededError`) if adding this file's `decrypted_size` to the aggregate `decrypted_size` of all in-progress (`state="init"`) uploads for the box would exceed the box's assigned cap.
```
Cito: Maybe 507 would be better than 409?

TheByronHimes: I guess that is another possibility

88-pistol-shrimp/technical_specification.md
TheByronHimes: The TTL here is for the life of an entire file upload, which is longer than a single presigned url ttl. Or maybe I misunderstand what you mean. I should update the wording too, since session already has another meaning.

**Condition matching:**  
I wish.
I spent an entire afternoon trying to make that work, but the bottom line seems to be that multipart uploads only work with presigned PUT urls, not presigned POSTs. Presigned POSTs give you fine control through headers like `Content-Length-Range`, but this is totally absent in presigned PUTs. This [article](https://advancedweb.hu/differences-between-put-and-post-s3-signed-urls/) is one of the clearest sources of documentation on the distinction, which is woefully absent in the AWS, botocore, and boto3 docs. I tried to manually create presigned URLs with the content headers, but didn't get it to work, so I'm not sure if it's actually possible to do it that way or not. No matter how you try to smuggle in the headers S3 seems to ignore it.

88-pistol-shrimp/technical_specification.md
TheByronHimes: I've changed the phrasing to be clearer now - did that address your concern?

88-pistol-shrimp/technical_specification.md
TheByronHimes: Yes, we can do that with optimistic locking.

88-pistol-shrimp/technical_specification.md
TheByronHimes: Probably 400

88-pistol-shrimp/technical_specification.md
Cito: Would be good to document this in the spec for the records if somebody wonders.

TheByronHimes: Done in [80393f2](https://github.com/ghga-de/epic-docs/pull/166/commits/80393f246911ae28490de2b2fde18c3bba88d799)

Cito (approved):

PR #167, opened 2026-04-30
Metadata ecosystem updates

sbilge:


88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+2. **Metldata transformation additions** — the `duplicate_class` transformation is retained but updated to raise an error when the input schema enforces globally unique IDs, since duplicating a class would produce ID collisions; the new `add_class` transformation is added to support building aggregate transformation workflows.
+3. **Metldata workflow runner API** — the internal `TransformationHandler` loop is encapsulated behind a clean `WorkflowRunner` public class that separates model derivation from data transformation, eliminating the API leakage that currently forces third-party services to manage low-level transformation details.
+4. **em-transformation-service updates** — the service adopts `WorkflowRunner` to replace its manual transformation loop, and its aggregate workflow configuration is rewritten using the new GUID-compatible transformations.
+
+### Terminology
+
+`Global Unique ID (GUID)`: A resource identifier that is unique not just within a single class but across all classes in a schemapack schema. GHGA accessions are an example: every Study, Sample, Dataset, etc. receives a globally unique accession.
```
mephenor: Looking at 
```
1. **Schemapack GUID support** — an opt-in `global_unique_ids` flag ensures resource IDs are unique across all classes in a datapack.
```
above, this should either refer to datapack here or be updated to reference schemapack above.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+`Model Derivation`: Running a workflow's transformation steps on an input schema to compute the output schema. This is done at configuration/startup time, once, independently of any data.
```
mephenor: Slightly imprecise, the config entities are also data that is persisted.
Would rephrase this from `data` to `datapacks`

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+`Data Transformation`: Running a workflow's transformation steps on a datapack to produce a transformed datapack. This is done at runtime, once per incoming datapack.
```
mephenor:
```suggestion
`Data Transformation`: Running a workflow's transformation steps on a datapack to produce a transformed datapack. This is done at runtime.
```

Reprocessing without changes is possible due to config changes.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+`Transformation Registry`: A mapping from transformation names (strings) to their `TransformationDefinition` objects. It is the authoritative lookup table used when resolving workflow step names to executable transformations.
```
mephenor:
```suggestion
`Transformation Registry`: A mapping from transformation names to their `TransformationDefinition` objects. It is the authoritative lookup table used when resolving workflow step names to executable transformations.
```

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+`WorkflowRunner`: The new high-level public class in metldata that executes a workflow. It exposes `transform_model` and `transform_data` as independent operations, hiding the internal `TransformationHandler` loop.
+
+`Aggregate Workflow`: A metldata workflow that produces differently-structured representation of an ingress schema — for example, transforming a relational Experimental Metadata Ingress Model (EMIM) into a Universal Discovery Model (UDM). Requires the ability to introduce new classes (`add_class`) in addition to the existing content-and-relation transformations.
```
mephenor: Needs some clarification here:
Is this a unique aggregate workflow?
What is aggregated?

In my understanding every workflow that performs more than one operation could be called an aggregate workflow.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+`Universal Discovery Model (UDM)`: The common target schema to which all EMIMs are transformed, serving as the basis for data queries and data display in the GHGA Data Portal.
```
mephenor: Does basis here mean source of truth for downstream?

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+The flag defaults to `false`, preserving backwards compatibility with all existing schemapacks and datapacks.
+
+##### Datapack validation
+
+When validating a datapack against a schema with `global_unique_ids: true`:
+
+1. Collect all resource IDs across all classes into a single flat list.
+2. Detect duplicates.
+3. If any ID appears in more than one class, raise a validation error that names the conflicting ID and every class in which it appears.
```
mephenor: If we need to track the IDs anyway, then we can implement this as a dict and (`resource_id`: [class_name_1, class_name_2, ...]) entries, filtering for everything with more than one value for duplicate detection.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+##### Serialization fix (frozen dicts / circular reference)
+
+Schemapack's serialization of `DataPack` objects (which use frozen dicts internally) currently triggers `ValueError: Circular reference detected (id repeated)` in naive client code. The fix shall be applied inside schemapack so that client code requires no workarounds.
```
mephenor: This could point to the existing analogous Schempack solution

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+#### 2. Metldata: `duplicate_class` Compatibility with Global Unique IDs
+
+The `duplicate_class` transformation shall be retained without a per-step model-assumption check for `global_unique_ids`. Adding such a check at the `duplicate_class` step would be too strict: a workflow that duplicates a class and then deletes it in a later step produces no ID collision in the final output, yet a per-step check would reject it.
```
mephenor: This is phrased in somewhat cumbersome way.
This can be simplified to:
Transient violation during workflow execution is allowed for duplicate_class inside a workflow, deferring the check to the overall validation logic. Only the final workflow result has to conform to the `global_unique_ids` constraint.

Needs some rephrasing and maybe merging with the next paragraph, as that one is much clearer on what's going on.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+##### Configuration model
+
+```python
+class RelationDefinition(BaseModel):
+    target_class: str
+    mandatory_origin: bool = False
+    mandatory_target: bool = False
+    multiple_origin: bool = False
+    multiple_target: bool = False
+    description: str | None = None
```
mephenor: Doesn't this already exist?

If so, mark it as already existing and move it below the `AddClassConfig` as it makes more sense for that to directly follow the last paragraph

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+class AddClassConfig(BaseModel):
+    class_name: str                                    # PascalCase; must not already exist in the schema
+    id_property_name: str                              # the property used as the resource ID
+    content_schema: dict                               # a valid JSON Schema object for the class content
+    relations: dict[str, RelationDefinition] | None = None  # optional; default no relations
```
mephenor:
```suggestion
    relations: dict[str, RelationDefinition] | None = None
```

The default is obvious without the description.

Are the relations modeled that way internally, i.e. with the None?
If not, then an empty dict would also do the job for the default.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+    description: str | None = None
+```
+
+Relations are declared at class-creation time when they are known upfront. Each key is the relation name (snake_case); the value specifies the target class and multiplicity/mandatory flags matching the schemapack relation definition format. All referenced target classes must already exist in the schema — `add_class` raises `ModelAssumptionError` if any target class is missing.
+
+**How to populate this newly added class and the relations are a question mark. 
+**
```
mephenor: Can we have a corresponding `AddData` transformation?
Would it make sense to separately populate the resources and the relations?

We have and object content schema that we could theoretically use to dynamically generate the corresponding config class, though I don't think this is supported natively by Pydantic, so we'd either need to have our own custom logic around this or look for an existing, lightweight solution.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+    ) -> DataPack:
+        """
+        Run the data-transformation phase of every workflow step in order.
+        Derives the intermediate schemas internally (equivalent to transform_model
+        applied step-by-step) — callers do not manage schema state.
+        Validates the input data before the first step and the output data
+        after the last step. Returns the final transformed datapack.
+        """
+        ...
+```
```
mephenor: We still have the schema derivation in both transform functions.
If we need to compute that either way, we could think about caching the intermediates and output between both functions, indexed by step name and some unique key (some ID on the `input_schema`, maybe?) that can be reused between the two of them.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+    workflow: Workflow,
+    transformation_registry: TransformationRegistry | None = None,
+) -> None:
+    """
+    Raises WorkflowValidationError if any step name is not in the registry
+    or if step configurations are invalid. Uses the default registry when
+    transformation_registry is None.
+    """
+    ...
+```
```
mephenor: Is this new?

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+`WorkflowRunner` shall be importable from `metldata.workflow.runner`. The existing `WorkflowHandler` and `TransformationHandler` remain available internally but are not part of the public API.
```
mephenor: Is this the same module as the other public exports?

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+After the change, the service no longer imports `TransformationHandler`, manages intermediate schema state, or resolves transformation names from the registry directly. A `WorkflowRunner` instance is constructed once per route at config-validation time and reused for every data transformation on that route.
+
+#### 7. Metldata: Aggregate Workflow Integration Test
+
+An end-to-end integration test shall be added to metldata that exercises all the new capabilities introduced in this epic together: a `global_unique_ids` input schema, `add_class`, `duplicate_class` failing correctly on such a schema, and `WorkflowRunner` running model derivation and data transformation separately.
+
+The test uses a representative aggregate workflow — modelled on the EMIM → UDM transformation — and verifies:
+
+- **`global_unique_ids: true` schema.** The input schemapack has the flag set; validation correctly rejects datapacks with duplicate IDs across classes.
```
mephenor: Formatting is weird and makes this harder to read, please rephrase and format differently.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+
+#### 8. Metldata: Performance Benchmark of the Aggregate Workflow
+
+A performance benchmark shall be run against the full EMIM → UDM aggregate workflow using the Epignostix dataset as a representative real-world input. The goal is to establish a baseline transformation time before the workflow is deployed to production.
```
mephenor: This is a public dataset, right?

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+Results are recorded and attached to the epic. They inform whether any transformation steps are unexpectedly slow and whether further optimisation is needed before shipping. If any single step accounts for a disproportionate share of the total time, it is flagged for investigation.
```
mephenor: Do we have a test harness for that or are we going the profiling route here?

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+Concretely:
+
+1. Fork `IBM/jsonsubschema` to `ghga-de/jsonsubschema`.
+2. Apply the fix from [IBM/jsonsubschema@8e6535](https://github.com/IBM/jsonsubschema/commit/8e6535461d00f37e4c06189826d320e4750d3a31) on the fork.
+3. Tag and publish a release from the fork.
+4. Update metldata's dependency to reference the GHGA fork release.
+5. Identify the specific GHGA CV value that triggers the bug and add a regression test to metldata that fails without the fix and passes with it.
+
+
+**Another issue with the library.** The library only supports JSON Schema draft 4. The GHGA metadata schema declares `$schema: https://json-schema.org/draft/2019-09/schema`. Keywords introduced after draft 4 (e.g., `$defs`, `unevaluatedProperties`) are silently ignored or mishandled, making subset validation unreliable for schemas that use draft 2019-09 features. Will we tackle this issue somehow? or will we fix the version to draft 4 in our content schemas?
```
mephenor: Fixing to an older version if we need new features seems like opting into more technical debt long term.
How big is this library, who will be responsible for maintaining it?

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Not included
+
+- A `add_relation` or `delete_relation` transformation — not required for the current aggregate workflow. (or is it?)
+
+## Human Resource/Time Estimation
+
+Number of sprints required: 3
+
+Number of developers required: 2
```
mephenor, line 348: There are a few heterogeneous concerns in here, might make more sense to group them by topic and give estimates for the different topics.

mephenor (commented):
More of a high level overview, only skimmed over the user journeys.
Some of the content needs rephrasing/reordering or clarifications.

This also feels like a bit too much of a mixed bag to fit into one epic and needs at least as bit more logical grouping to further evaluate.

The jsonsubschema story feels like it could be it's own thing.
It's not really clear what the short and long term effort will be to fork, patch and maintain the fork.
Seems like there might be quite a bit of a hidden cost.
This needs further digging into or more details of what's already known.

88-mermaids-purse/technical_specification.md
sbilge: Yes, for the UI.

88-mermaids-purse/technical_specification.md
sbilge: Does it already exist where? It exists in schemapack library itself, not in metldata.

87-mermaids-purse/technical_specification.md
sbilge: I think it is okay this way. We will need to touch multiple libraries anyway.

88-mermaids-purse/technical_specification.md
sbilge: yes

88-mermaids-purse/technical_specification.md
sbilge: No, the name is very GHGA specific. 

This is the original aggregate stats workflow that we reproduce with the new metldata implementation:   https://github.com/ghga-de/metadata-config/blob/ec0f28fe4514e6d8dcdb16369345041a32ca9ece/configuration/metadata_config.yaml#L29

I repharased.

88-mermaids-purse/technical_specification.md
sbilge: profiling

88-mermaids-purse/technical_specification.md
sbilge: The second issue with the older JSON version support is open to discussion. Making our release with the bug fix that we need is not costly - especially when compared to writing such a library ourselves. 

We can consider utilising AI agents on this library.

88-mermaids-purse/technical_specification.md
sbilge: I don't think add_class is necessary for the aggregation stats workflow at all. I will initiate a discussion round.

88-mermaids-purse/technical_specification.md
sbilge: I want to explore if we can invoke them separately, but it seems difficult.

sbilge: After exploring how the model transformation is handled in `TransformationHandler`, I changed this section. The solution seems easier than we thought. The derived schema is already cached; it is not repeated every time we call `data_transform()`.

88-mermaids-purse/technical_specification.md
mephenor: Ok, so we can just pull that in where needed.
I was confused to see that here as a seemingly new thing.

Still, better for readability to move this below `AddClassConfig` as the text above directly talks about `add_class`.

88-mermaids-purse/technical_specification.md
mephenor: Might we need that outside of the aggregation, i.e. would there be a general use to have separate `AddClass` and `AddData` transformations?

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,309 @@
...
+Concretely:
+
+1. Fork `IBM/jsonsubschema` to `ghga-de/jsonsubschema`.
+2. Apply the fix from [IBM/jsonsubschema@8e6535](https://github.com/IBM/jsonsubschema/commit/8e6535461d00f37e4c06189826d320e4750d3a31) on the fork.
+3. Tag and publish a release from the fork.
+4. Update metldata's dependency to reference the GHGA fork release.
+5. Identify the specific GHGA CV value that triggers the bug and add a regression test to metldata that fails without the fix and passes with it.
+
+
+**Another issue with the library.** The library only supports JSON Schema draft 4. The GHGA metadata schema declares `$schema: https://json-schema.org/draft/2019-09/schema`. Keywords introduced after draft 4 (e.g., `$defs`, `unevaluatedProperties`) are silently ignored or mishandled, making subset validation unreliable for schemas that use draft 2019-09 features. We will not tackle this in the scope of this epic. But it is something to consider for the future usage of that library.
```
mephenor: This should be adapted to what we discussed on another channel to reflect current plans better.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,309 @@
...
+
+#### 2. Metldata
+
+##### 2.1. `duplicate_class` Compatibility with Global Unique IDs
+
+This section does not include any changes to `duplicate_class` itself, which continues to function as it currently does. It only clarifies how the presence of `global_unique_ids` in the schema affects the validation of workflow outputs that use `duplicate_class`.
+
+The `duplicate_class` transformation shall be retained in the transformation registry; the `global_unique_ids` constraint does not conflict with its presence. There are two scenarios to consider:
+1. The schema does not have `global_unique_ids` set. The `duplicate_class` transformation works as it currently does, with no issues.
+2. The schema has `global_unique_ids: true`. If a workflow duplicates a class and later deletes it, the final output contains no ID collision. Transient ID collisions introduced by `duplicate_class` within a workflow are permitted — the `global_unique_ids` constraint is enforced only on the final workflow output, not on intermediate steps.
```
mephenor: This is much easier to read, thank you

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,309 @@
...
+
+The second scenario is consistent with the existing metldata behaviour: when the no-intermediate-validation flag is set, model compatibility is not enforced after every step — only at the workflow boundaries. The `global_unique_ids` uniqueness constraint follows the same rule: it is enforced only at **workflow output validation** (the final step), not during intermediate steps. If the output schema of a workflow still has `global_unique_ids: true` and the output datapack contains duplicate IDs across classes, the post-workflow validation raises an error at that point.
+
+Two test cases should be added to ensure:
+1. When no-intermediate-validation flag is set, a workflow that uses `duplicate_class` and produces a final schema with `global_unique_ids: true` does not raise an error given that the output datapack has no ID collisions across classes.
+2. When intermediate validation is enabled, the same workflow raises an error at the step where `duplicate_class` is applied, since the intermediate schema at that point contains duplicate IDs across classes.
```
mephenor, line 94: More of an implementation detail, but the general failure case of just copying and leaving it as is should also be covered for the first test case.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,309 @@
...
+To bridge this constraint change between input and output schemas, the `replace_resource_ids` transformation must be extended with an optional argument specifying whether the replacement IDs are expected to be globally unique across classes:
+
+```yaml
+class_name: File
+global_unique_ids: true
+```
+
+This argument defaults to `false`, preserving compatibility with existing behaviour. When set to `true`, the transformation sets `global_unique_ids: true` on the output schemapack, signalling that the output schema enforces the globally-unique constraint.
+
+No additional validation logic is required: the post-transformation validation step will enforce the `global_unique_ids` constraint on the output schemapack and verify the transformed datapack against it.
```
mephenor: What do we do with existing collisions here, do the AEMPacks just fail validation at the transformation step?
This feels like this would warrant a special feedback mechanism.

With the current design of the EMTS this would not be caught at the phase where we can simply move it to the DLQ, i.e. currently we assume when transformations fail it's a service misconfiguration and let it crash.

With this case added, that behavior is wrong and we need some error handling and policy as this is transformation config dependent and can't be checked a priori.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,309 @@
...
+
+It would also be possible to set a different content schema for the new class via `add_class`, but the existing content transformation already covers that use case.
+
+One potentially missing primitive in `duplicate_class` is relation deletion. In the current aggregation workflow, the duplicated class inherits all relations from the source class. Those relations are removed indirectly when the classes that are no longer needed are deleted — their removal cascades to any references pointing at them, which clears the duplicated class's relations as a side effect. A `delete_relation` transformation would make this explicit, allowing a duplicated class's relations to be modified without having to delete other classes.
```
mephenor: Need some time to think about this.

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,309 @@
...
+1. The em-transformation-service starts up and loads its transformation configuration.
+2. For each route, it instantiates a `WorkflowRunner` and calls `runner.transform_model(input_schema)` to derive the output schema. No datapacks are required at this stage.
+3. The derived output schema is stored alongside the route and used to validate incoming data.
+4. When an AnnotatedEMPack arrives at runtime, the service calls `runner.transform_data(data, input_schema, annotation)` to produce the derived datapack. The service does not manage `TransformationHandler` instances or intermediate schema state.
+5. The resulting datapack is validated against the pre-derived output schema and published.
+
+## Additional Implementation Details
+
+- The `global_unique_ids` flag defaults to `false`. No existing schemapack file or datapack needs to be modified.
+- `WorkflowRunner.transform_data` must internally re-derive the intermediate schemas at each step — it cannot assume that `transform_model` has been called first. This keeps the two methods fully independent.
```
mephenor: Does this still hold for the updated design?

88-mermaids-purse/technical_specification.md
sbilge: a wrapper is created around `validate_workflow_against_registry`

88-mermaids-purse/technical_specification.md
sbilge: I am really not sure about this add_class transformation as a whole. I will change this part after our scheduled discussion.

88-mermaids-purse/technical_specification.md
sbilge: I reworked this part.

88-mermaids-purse/technical_specification.md
sbilge: partially. but I changed this part, too.

87-mermaids-purse/technical_specification.md
```diff
@@ -11,45 +11,44 @@ Epic planning and implementation follow the
...
 
-`WorkflowRunner`: The new high-level public class in metldata that executes a workflow. It exposes `transform_model` and `transform_data` as independent operations, hiding the internal `TransformationHandler` loop.
+`WorkflowRunner`: The new high-level public class in metldata that executes a workflow. It accepts `workflow` and `input_model` at construction, eagerly derives the output schema (exposed via `.model`), and defers data transformation to `.run_workflow(data, annotation)` calls.
 
-`Aggregate Stats Workflow`: A GHGA-specific metldata workflow that produces another representation of the GHGA ingress model that is optimised for aggregate statistics queries. 
+`Aggregate Stats Workflow`: A GHGA-specific metldata workflow that produces a representation of the GHGA ingress model optimised for aggregate statistics queries.
 
-`Universal Discovery Model (UDM)`: The common target schema to which all EMIMs are transformed, serving as the basis for data queries and data display in the GHGA Data Portal.
+`Universal Discovery Model (UDM)`: The common target schema to which all EMIMs are transformed, serving as the basis for data queries and display in the GHGA Data Portal.
```
mephenor, line 38: PAM is used in the data transformation section, but not defined here

88-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,457 @@
...
+
+This transformation walks the flat ID listing on a source class, resolves each ID to the file class it belongs to, and writes a relation pointing back to the source class on the resolved resource. It differs from infer_relation in that the source of the link information is a content property holding resource IDs, not an existing relation path.
+
+**Assumptions**
+
+The source class declared in the configuration must exist in the schema and must declare the content property holding the ID listing.
+The candidate target classes declared in the configuration must exist in the schema.
+The relation name declared in the configuration must not already exist on any of the candidate target classes.
+Each resource ID held by the source content property must resolve to a resource in exactly one of the candidate target classes. An unresolvable ID, or an ID present in more than one candidate class, is a data transformation error.
```
mephenor: Can be converted to a list for readability.

87-mermaids-purse/technical_specification.md
```diff
@@ -62,92 +61,79 @@ classes:
...
 
-The second scenario is consistent with the existing metldata behaviour: when the no-intermediate-validation flag is set, model compatibility is not enforced after every step — only at the workflow boundaries. The `global_unique_ids` uniqueness constraint follows the same rule: it is enforced only at **workflow output validation** (the final step), not during intermediate steps. If the output schema of a workflow still has `global_unique_ids: true` and the output datapack contains duplicate IDs across classes, the post-workflow validation raises an error at that point.
+The second scenario is consistent with the existing metldata behaviour: when the no-intermediate-validation flag is set, model compatibility is not enforced after every step — only at the workflow boundaries. The `globallyUniqueIds` uniqueness constraint follows the same rule: it is enforced only at **workflow output validation** (the final step), not during intermediate steps. If the output schema of a workflow still has `globallyUniqueIds: true` and the output datapack contains duplicate IDs across classes, the post-workflow validation raises an error at that point.
```
mephenor, line 90: -> If the output schema of a workflow ~~still~~ has `globallyUniqueIds: true`

88-mermaids-purse/technical_specification.md
```diff
@@ -62,92 +61,79 @@ classes:
...
+The need for the `add_class` transformation comes from the new design of the metadata schema, which removed administrative metadata from the current schema. The classes removed are `Study`, `Publication`, `Dataset`, `DataAccessPolicy`, and `DataAccessCommittee`. This also results in removing the `dataset` property from the file-related classes: `IndividualSupportingFile`, `ExperimentMethodSupportingFile`, `ResearchDataFile`, `AnalysisMethodSupportingFile`, and `ProcessDataFile`.
```
mephenor: Could you convert the inline enumerations into lists here?

88-mermaids-purse/technical_specification.md
```diff
@@ -157,14 +143,194 @@ Relations are declared at class-creation time when they are known upfront. Each
...
+    "type": "object"
+  }
+relations:
+  - relation_name: study
+    description: "A relation from Dataset to Study."
+    target_class: Study
+    target_resource_id_field: study_id
+    mandatory: MandatoryRelationSpec
+    multiple: MultipleRelationSpec
+```
```
mephenor: This is quite a text wall with regard to complexity introduced by sheer amount of different terms.

Inline enumerations could be converted to lists for some readability improvements, but this could use a small diagram that's a bit more generic about the different entities involved to better display the process.
This could include parts of the config/schema/datapack with the irrelevant parts stripped out.

88-mermaids-purse/technical_specification.md
```diff
@@ -157,14 +143,194 @@ Relations are declared at class-creation time when they are known upfront. Each
...
+**Open questions:**
+- The PAM annotation also carries `created: Date`, `changed: Date`, and `active: bool`. Should these be included in the content schema, or can they be ignored since they are not needed for the aggregate stats workflow?
```
mephenor: Artifacts published downstream should be tailored to their consumers, so in general they could be dropped at some point in the transformation pipeline if not needed downstream.

We could propagate the information to the UDM and then drop it there for downstream artifacts that don't need it, while keeping the possibility open for newer workflow changes to consume it from there.

Though not really sure if there's a use case for that, so needs a bit more thinking about

88-mermaids-purse/technical_specification.md
```diff
@@ -157,14 +143,194 @@ Relations are declared at class-creation time when they are known upfront. Each
...
+- Should DAC and DAP be received as separate classes or inlined into the Dataset content schema? (The example above uses inline.)
```
mephenor: Not really qualified to comment on that.
What are the implications?

88-mermaids-purse/technical_specification.md
```diff
@@ -180,126 +346,108 @@ for step in workflow.workflow.operations:
...
+On top of `WorkflowHandler`, a new public `WorkflowRunner` class is introduced in `src/metldata/workflow/run.py`. It is registry-aware and accepts `workflow` and `input_model` as keyword arguments at construction. Model derivation happens eagerly in `WorkflowRunner.__init__`; the schema is exposed via the read-only `.model` property. Data transformation is deferred: callers invoke `.run_workflow(data=..., annotation=...)` once per datapack. This allows consumers to pay the schema-derivation cost once at construction and call `.run_workflow()` repeatedly without managing `TransformationHandler`, the registry, or intermediate schema state directly.
 
-The test uses a representative aggregate stats workflow and verifies that `global_unique_ids: true` is enforced: the input schemapack has the flag set, and validation correctly rejects datapacks with duplicate IDs across classes.
+A convenience function `validate_workflow()` is added to `src/metldata/workflow/validate.py`; it delegates to `validate_workflow_against_registry` using the built-in registry so callers do not need to pass a registry explicitly.
```
mephenor: This is quite low level and busy. 
Some of the details feel like they don't need to be fixed here, but could be left to the implementer.

No need to change the detail level, though, but grouping the changes as bullet point list per module/class could improve the structure here, i.e. easier to grasp without rereading the whole thing.

88-mermaids-purse/technical_specification.md
mephenor: Reposetd from another communication channel:

What about making that pull and not push, i.e. just have another collection that is populated  with touples of (PID, id+model_name of the AEMPack at derivation step that failed, actionable error description) and make that queryable from the outside?
For those cases we likely need a new version of the AEMPack either way, so some frontend functionality could prominently display some kind of `Action required` when checking the status of the associated study.

We need an API for the available models/schemas either way, so that would just need a separate endpoint

88-mermaids-purse/technical_specification.md
sbilge: Most of these "text wall" are the configurations that will be used in the aggregation stats workflow. It does not depict a concept or a process flow that can be shown with a diagram. I created a separate appendix section and provided configs there.

88-mermaids-purse/technical_specification.md
sbilge: it is not very fundamental to the changes in the libraries. The aggregation workflow can be tweaked easily once we know the requirements of the downstream consumer.

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+- Add the new `ClassDefinition` (with content schema and any declared relations) to the schemapack.
+- Raise `ModelAssumptionError` if a class with `class_name` already exists.
+- Raise `ModelAssumptionError` if any `target_class` in `relations` does not exist in the schema.
+
+**Data transformation**
+
+For the class declared in `config.class_name`, the transformation inserts the resources sourced from the PAM into the DataPack. Each PAM resource is processed individually and split into two parts before being written:
+
+- **Content** — properties that match the JSON Schema declared in `config.content_schema` are kept under the resource's content.
+- **Relations** — for every entry in `config.relations`, the property named by `target_resource_id_field` is read from the incoming PAM record, removed from content, and written as a relation under `relations.<relation_name>` with `targetClass = target_class` and `targetResources` set to the value(s) read.
```
mephenor, line 160: How would this splitting/matching work in practice?
Deal with the relations first and try to fit the content in afterwards and reject if it doesn't fit the `content_schema`?

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+
+**Model transformation**
+
+- Add the new `ClassDefinition` (with content schema and any declared relations) to the schemapack.
+- Raise `ModelAssumptionError` if a class with `class_name` already exists.
+- Raise `ModelAssumptionError` if any `target_class` in `relations` does not exist in the schema.
+
+**Data transformation**
+
+For the class declared in `config.class_name`, the transformation inserts the resources sourced from the PAM into the DataPack. Each PAM resource is processed individually and split into two parts before being written:
```
mephenor, line 157: The following section implies sort of a  schema for the data contained in the annotation.

Currently the annotation is just a dict.
It would probably make sense to define that via Pydantic and induce some structure within specific parts of the annotation that way.

Another point to consider: The data transformation step is described in terms of PAM data extraction/population, but the interface should be generic enough to not be bound to the data source.

Just some thoughts, nothing that needs to be addressed directly in the specs here.

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+The transformation first builds a global mapping from each ID listed under `source_content_property` to the set of source resources that reference it. Iterating `resources.<source_class_name>` once produces:
+
+```
+file_id_1 → [DS_A, DS_B]
+file_id_2 → [DS_A]
+file_id_3 → [DS_B, DS_C]
+...
+```
+
+- The keys of this mapping are intersected with the resource IDs present under `resources.<class_name>` for the class being processed. For each ID in the intersection, the corresponding set of source resource IDs is written to `relations.<relation_name>.targetResources` on that resource.
```
mephenor: This is still a bit hard to read.

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+- IDs outside the intersection belong to other classes and are left untouched — they will be picked up by their own invocations.
```
mephenor: What exactly does that mean?
The mapping is created once, but we iterate over each target class and look at the at the relevant IDs for the class?
What happens with leftovers, i.e. does this need to be exhaustive?

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+| Name                | Kind     | Purpose                                                     |
+| ------------------- | -------- | ----------------------------------------------------------- |
+| `WorkflowRunner`    | class    | High-level workflow execution; `.model` + `.run_workflow()` |
+| `validate_workflow` | function | Validates a workflow against the built-in registry          |
+
+
+
+##### 2.7. Aggregate Stats Workflow Integration Test
+
+An end-to-end integration test shall be added to metldata, exercising all new capabilities together: a `globallyUniqueIds` schema, `add_class` (pending resolution of open questions in section 2.5), correct error behaviour for `duplicate_class` when intermediate validation is enabled, and `WorkflowRunner` running model derivation and data transformation independently.
```
mephenor: Is the `(pending resolution of open questions in section 2.5)` part still relevant?

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+
+The test uses a representative aggregate stats workflow and verifies that the `globallyUniqueIds` constraint is enforced on the output: validation rejects datapacks with duplicate IDs across classes.
+
+
+##### 2.8. Performance Benchmark of the Aggregate Stats Workflow
+
+A performance benchmark shall be run against the full EMIM → aggregate stats workflow using the Epignostix dataset as a representative input. Profiling is done via `cProfile`. The goal is to establish a baseline transformation time before production deployment. Results are attached to the epic. Any step that accounts for a disproportionate share of total time is flagged for investigation.
```
mephenor:
```suggestion
A performance benchmark shall be run against the full EMIM -> aggregate stats workflow using the Epignostix dataset as a representative input. Profiling is done via `cProfile`. The goal is to establish a baseline transformation time before production deployment. Results are attached to the epic. Any step that accounts for a disproportionate share of total time is flagged for investigation.
```

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Journey 1: Schema Designer Enables Globally Unique IDs
+
+1. A schema designer adds `globallyUniqueIds: true` to the schema YAML.
+2. Running `schemapack validate` against a conforming datapack succeeds only if all resource IDs are distinct across all classes.
+3. If a collision exists — e.g., both a Study and a Sample have the ID `GHGAS00001` — validation fails with a clear error naming the conflicting ID and both classes.
+
+### Journey 2: Workflow Author Builds an Aggregate Stats Workflow with `add_class`
+
+1. A workflow author designing the EMIM → UDM workflow needs to reintroduce administrative classes (e.g., `Study`, `Dataset`) that are absent from the experimental EMIM but present in PAM annotations.
```
mephenor:
```suggestion
1. A workflow author designing the EMIM -> UDM workflow needs to reintroduce administrative classes (e.g., `Study`, `Dataset`) that are absent from the experimental EMIM but present in PAM annotations.
```

87-mermaids-purse/technical_specification.md
```diff
@@ -0,0 +1,487 @@
...
+- IDs outside the intersection belong to other classes and are left untouched — they will be picked up by their own invocations.
+
+##### 2.5. Aggregate Stats Workflow: Adding Back PAM Classes via `add_class`
+
+The classes that will be added back to the EMIM via the aggregate stats workflow are:
+1. `Study`,
+2. `Publication`,
+3. `Dataset`.
+
+`Dataset` will have `DataAccessCommittee` and `DataAccessPolicy` inlined, so they will not appear as separate classes in the intermediate or final schema. The full `add_class` configurations for these classes are in the [Appendix](#appendix-pam-transformation-configurations).
```
mephenor, line 218: Could list the property name for the inlined classes here, i.e. `dac` and `dap`

mephenor (changes requested):
Much easier to read now, just needs some minor tweaks and clarifications.

87-mermaids-purse/technical_specification.md
sbilge: They are provided in the appendix section. The configuration for the `Dataset` has dap and dac inlined.

87-mermaids-purse/technical_specification.md
sbilge: During the data transform, we will have access to all information necessary. We don't have to have a hard separation on which one to implement first. The details can be figured out during the implementation.

87-mermaids-purse/technical_specification.md
sbilge: There won't be leftovers on the global mapping. All the transformation does is invert the relation. If a dataset refers to a file, that file will refer to that dataset.

87-mermaids-purse/technical_specification.md
mephenor: Fair enough

87-mermaids-purse/technical_specification.md
mephenor: Ok, let's move that to the implementation.
I feel like that's the part that contains a few hidden gotchas, so probably will need some additional care.

mephenor (approved):

PR #168, opened 2026-04-30
DAO batch methods

mephenor:


90-red-billed-quelea/technical_specification.md
```diff
@@ -0,0 +1,107 @@
...
+### Index updating/recreation
+
+Currently `_get_dao` on the provider calls `collection.create_index` without any error handling. 
+If a different index is provided after initial creation, this will fail due to mismatching fields on the index.
+This can be solved by migrations, but to also support dynamically updating TTL indices, either dropping and recreating or using the `collMod` command to modify the index in place would be a preferable option.
+This needs a small bit of investigation to check, if `collMod` is even supported for non-TTL indices and if so, which of the two options is the better/more general one.
+
+## Implementation Details
+
+New implementations for the protocol methods are added to `MongoDbDao` in and reuse the existing `translate_pymongo_errors` context manager together with `_dto_to_document` and `_document_to_dto`.
```
mephenor:
```suggestion
New implementations for the protocol methods are added to `MongoDbDao` and reuse the existing `translate_pymongo_errors` context manager together with `_dto_to_document` and `_document_to_dto`.
```

90-red-billed-quelea/technical_specification.md
```diff
@@ -0,0 +1,107 @@
...
+## Scope
+
+### Outline:
+
+This epic aims to extend the `Dao` protocol in hexkit with batch methods that apply the existing  CRUD operations to several resources in a single call. 
+The corresponding MongoDB provider will implement those methods and the in-memory DAO testing utility will be updated accordingly.
+
+This will allow services to perform common bulk operations using one database round-trip per resource batch instead of one round-trip per resource.
+
+In addition, the existing index creation will be extended to allow for dynamic updates after creation.
```
TheByronHimes: I'm not sold on this yet. I think changes in the index should be performed through a DB migration because they inherently link two different versions of the application.

90-red-billed-quelea/technical_specification.md
```diff
@@ -0,0 +1,107 @@
...
+The existing single-resource methods remain unchanged and are not re-implemented in terms of the batch methods, though nothing precludes using the new methods with single item lists to achieve compatible behavior.
+
+An additional `BatchOperationError` will be introduced to wrap single failures within a batch operation, holding a dictionary of document IDs to error details, for handling by the caller.
+
+### Index updating/recreation
+
+Currently `_get_dao` on the provider calls `collection.create_index` without any error handling. 
+If a different index is provided after initial creation, this will fail due to mismatching fields on the index.
+This can be solved by migrations, but to also support dynamically updating TTL indices, either dropping and recreating or using the `collMod` command to modify the index in place would be a preferable option.
+This needs a small bit of investigation to check, if `collMod` is even supported for non-TTL indices and if so, which of the two options is the better/more general one.
```
TheByronHimes: I did some looking into this. collMod does not seem to be supported or recommended for managing indexes generally. It's used to modify TTL indexes, but for other index types we should drop and re-create. This is only exposed in pymongo via the [database command](https://pymongo.readthedocs.io/en/stable/api/pymongo/database.html#pymongo.database.Database.command) method, which we could perhaps plop [here](https://github.com/ghga-de/hexkit/blob/196625b86905357684ff472a7aefdc8fa004bd7e/src/hexkit/providers/mongodb/provider/dao.py#L459) on a conditional basis.

The MongoDB docs AI also said that the TTL-monitor is disabled in Standalone, but it wasn't clear if that was always the case or only if it's configured in a given way.

mephenor: So drop and recreate it is, thank you.

I have to have a look at the TTL monitor thing.
The way I read it, in a replica set only the primary has the TTL monitor running, but if you move a node out of a replica set to standalone, it no longer works for that node.
Not really sure what that implies about a standalone setup in general, the docs are not very clear about that.

90-red-billed-quelea/technical_specification.md
mephenor: Maybe it would make sense to treat the normal and TTL index as separate concepts and expose them that way?

We could expose drop and recreate logic, but limited to manual use in migrations and automatic recreation in the TTL case?

90-red-billed-quelea/technical_specification.md
TheByronHimes: That was almost verbatim what I read, so I think we're on the same page.

90-red-billed-quelea/technical_specification.md
TheByronHimes: I'm okay with that. Conditionally call `self._db.command({"collMod": name, "index": ...})` if the index is TTL

90-red-billed-quelea/technical_specification.md
mephenor: Still trying to figure out what the actual states would be where the TTL monitor is disabled.
From some further exploration, that note did not seem to exist in previous versions of the docs in this form.

I also don't like that this is a potential footgun that couples code concerns with ops issues.
Maybe it's a bad idea to expose this via hexkit and we should fall back to direct pymongo manipulation where this is needed to not artificially bloat the problem surface here.

If we do index handling via drop & recreate, I'll just focus this epic on the batch operations and drop the index related stuff.

TheByronHimes (approved):

